PR #6, opened 2022-08-17
Run only tests + docs on PR or workflow dispatch

nabriis:
Currently the docs are compiled both on PR and on push. Such that if one pushes and creates a PR the docs are compiled twice.

Instead, this sets it up so docs and tests are only run when a PR is created or updated.  It is also possible to manually run tests/docs using the workflow dispatch.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,5 +1,5 @@
 name: docs
-on: [push, pull_request, workflow_dispatch]
+on: [pull_request, workflow_dispatch]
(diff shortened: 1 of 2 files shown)


amal-ghamdi (approved):
Thank you @nabriis 

I just took a quick look at these events:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows

I think this is a reasonable set up

jakobsj (approved):
Sounds good to me.

PR #8, opened 2022-08-17
Revert ABC Distribution and Sampler back to their modules (removing cuqi.core)

### LINKED ISSUE #7: Revert core module
Splitting CUQIpy into a core module and extra auxiliary methods such as specific distributions, samplers etc. does not seem to be feasible in the long run. Hence this PR reverts those recent changes.

The main reason for this is that much of the logic is based on specific classes like the Gaussian distribution, Continous1D geometry etc. I would suggest that we instead think of making CUQIpy as a whole the "core" package and use the plugin structure to add extra features that do not belong to the core.

A particular point of interest is to, in the long term, make the core array agnostic. That is, have an array api for CUQIpy that can use numpy, pytorch etc. as backend. This is also mentioned in previous issues on GitLab. Making CUQIpy array agnostic will go a long way to making the package more easily extended in the future. For example, every current distribution in the "core" depends on NumPy and SciPy, but (perhaps) fairly easily could be written such that any array API following the [NEP 47](https://numpy.org/neps/nep-0047-array-api-standard.html) standard would work. The same is true for samplers (NumPy-based) and Geometry (Matplotlib-based).

nabriis:
Closes #7


jakobsj:
Let's discuss this tomorrow.

amal-ghamdi (approved):
Thank you @nabriis 

I think removing core is more clear from developer point of view. But the thing that is nice about the core is for future development when different backend are used. It is nice to have the template in one place then. But may be it is too early for that now.

jakobsj (approved):

PR #11, opened 2022-08-19
Sprint10 plot ci func

### LINKED ISSUE #10: Add option to compute ci for function field
Figure1 shows plot_ci (plot credible interval) for KLExpantion type samples. Figure2 shows plot of the samples themselves. The discrepancy between how the ci looks and how the samples look is because for computing the ci, the ci is computed on the coefficient and then converted to function values (par2fun). We want to add an option for plot_ci to compute the ci for the function values rather than for the coefficients (Figure3).

Image
Figure1
Image
Figure2
Image
Figure3

amal-ghamdi:
closes #10.


cuqi/samples/_samples.py
```diff
@@ -190,6 +192,16 @@ def geometry(self):
...
 
+    @property
+    def _funvals(self):
+        """Return function values of the samples by applying :meth:`self.geometry.par2fun`"""
+        if not hasattr(self, '_samples_funvals'):
+            funvals = np.empty((self.geometry.dim, self.Ns))
+            for i, s in enumerate(self):
+                funvals[:, i] = self.geometry.par2fun(s)
+            self._samples_funvals = funvals
+        return self._samples_funvals
```
nabriis, line 203: Really good to cache the funvals in case they are used multiple times :).

However, if self.samples changes the funvals should also change. Perhaps adding a setter to self.samples that simply resets _samples_funvals to None or something.

cuqi/samples/_samples.py
```diff
@@ -190,6 +192,16 @@ def geometry(self):
             self._geometry = _DefaultGeometry(grid=np.prod(self.samples.shape[:-1]))
         return self._geometry
 
+    @property
+    def _funvals(self):
```
nabriis, line 196: Perhaps we should expose this property publicly? Perhaps make it a method and call it compute_funvals().

cuqi/samples/_samples.py
```diff
@@ -348,7 +362,7 @@ def hist_chain(self,variable_indices,*args,**kwargs):
         plt.legend(variables)
         return patches
 
-    def plot_ci(self,percent=95,exact=None,*args,plot_envelope_kwargs={},**kwargs):
+    def plot_ci(self, percent=95, exact=None, compute_on_par=True, *args, plot_envelope_kwargs={}, **kwargs):
```
nabriis, line 365: While we are at it, could you change plot_envelope_kwargs to None as default to avoid this issue with mutable defaults?

nabriis (changes requested):
Hi @amal-ghamdi. This PR looks really good. I like the new options it provides to the users. I had a small suggestion related to the caching of the function value samples.

cuqi/samples/_samples.py
amal-ghamdi: That is a great point. I made few updates to address this including creating samples property and setter (Now samples are not writable through slicing, but the entire samples array can be set to a new one.). This is to prevent the user from writing something like: `s1.samples[0,0]=1`, (where the code as is will not be able to detect this change and reset funvals.). I added some tests for the added property too.

cuqi/samples/_samples.py
amal-ghamdi: Great, just did that.

cuqi/samples/_samples.py
amal-ghamdi: Good point!

amal-ghamdi:
Thank you @nabriis for your helpful review! I made updates to address the points you raised.

tests/test_samples.py
```diff
@@ -125,18 +125,106 @@ def test_ess():
...
+def test_compute_ci(percent, compute_on_par, geometry):
     dist = cuqi.distribution.DistributionGallery("CalSom91")
     sampler = cuqi.sampler.MetropolisHastings(dist)
     samples = sampler.sample_adapt(500)
-    ci = samples.compute_ci(percent)
+    samples.geometry = geometry
+    ci = samples.compute_ci(percent, compute_on_par=compute_on_par)
+
+    if not compute_on_par:
+        samples = samples._funvals
```
nabriis, line 158:
```suggestion
        samples = samples.funvals
```

nabriis (approved):
Thanks Amal. Only a minor edit and you can pass it to Jakj :)

cuqi/samples/_samples.py
nabriis: Good with the locking of arrays. Did not think of that. It looks like a lot of code, but seems to be necessary.

jakobsj (commented):
Thank you @amal-ghamdi for this nice addition, being able to generate plots, compute mean etc on both function and parameter values will surely come in handy. My main comment is on how to trigger this behavior, through the flag as proposed or if there is a simpler way, as it seems slightly cumbersome to carry around this `compute_on_par` flag. Let us discuss shortly

cuqi/samples/_samples.py
```diff
@@ -190,6 +193,33 @@ def geometry(self):
...
         return self._geometry
 
+    @property
+    def samples(self):
+        return self._samples
+
+    @samples.setter
+    def samples(self, value):
+        # Disable write flag to prevent updates to the samples through slicing.
+        value.setflags(write=False)
```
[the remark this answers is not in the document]
amal-ghamdi: in the updated PR#23 we do not need samples setter and getter so this logic is removed.

cuqi/samples/_samples.py
```diff
@@ -190,6 +193,33 @@ def geometry(self):
...
+            self._funvals = None
+        if hasattr(self, '_geometry') and self._geometry is not None:
+            warnings.warn("Setting the samples value will reset the geometry.", UserWarning)
+            self._geometry = None
+
+    @property
+    def funvals(self):
+        """Return function values of the samples by applying :meth:`self.geometry.par2fun`"""
+        if not hasattr(self, '_funvals') or self._funvals is None:
+            self._funvals = np.empty((self.geometry.dim, self.Ns))
```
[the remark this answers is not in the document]
amal-ghamdi: I agree, I created issue issue #26  to address the performance here and for geometry plotting loops

cuqi/samples/_samples.py
```diff
@@ -287,11 +318,12 @@ def variance(self):
         """Compute pointwise variance of the samples"""
         return np.var(self.samples, axis=-1)
 
-    def compute_ci(self, percent = 95):
-        """Compute pointwise credibility intervals of the samples"""
+    def compute_ci(self, percent=95, compute_on_par=True):
```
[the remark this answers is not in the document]
amal-ghamdi: Yes I agree with you. That is a great point, now in #23 we removed the flag and use the behavior 

```python
samples.funvals.plot_ci(95)
```

amal-ghamdi:
Thank you @jakobsj for your review! and helpful feedback. I created a new PR #23 with the other implementation we discussed (which is to create a new  samples object and compute statistics on that one.) I will close this one to avoid confusion.

PR #15, opened 2022-08-19
Sprint10 step expansion

### LINKED ISSUE #14: Allow StepExpansion geometry to accept parameter n_steps
Allow StepExpansion geometry to accept number of steps as a parameter

amal-ghamdi:
closes #14


tests/test_geometry.py
```diff
@@ -130,3 +130,20 @@ def test_mapped_geometry(geom, map, imap):
...
+        with pytest.raises(ValueError):
+            cuqi.geometry.StepExpansion(grid,n_steps)
+    else:
+	#Otherwise, assert that the StepExpansion is correct
+        geom = cuqi.geometry.StepExpansion(grid,n_steps)
+        par = np.random.randn(n_steps)
+        geom.plot(par,linestyle = '', marker='.')
+
+        assert np.allclose(par, geom.fun2par(geom.par2fun(par))) \
+           and geom.dim == n_steps
```
amal-ghamdi, line 149:
```suggestion
           and geom.dim == n_steps
           
```

tests/test_geometry.py
```diff
@@ -130,3 +130,21 @@ def test_mapped_geometry(geom, map, imap):
...
+
```
amal-ghamdi, line 150:
```suggestion
```

cuqi/geometry/_geometry.py
```diff
@@ -703,31 +703,49 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
 class StepExpansion(Continuous1D):
     '''
     class representation of the step random field with 3 intermidiate steps
+
+    Parameters:
+    -----------
+    grid: ndarray
+        grid points
```
nabriis, line 710:
```suggestion
        Grid points for the step expansion to be evaluated at. There can be more grid points than number of steps.
```

cuqi/geometry/_geometry.py
```diff
@@ -703,31 +703,49 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
+
+    nsteps: int
+        number of equidistant steps
```
nabriis, line 713:
```suggestion
    n_steps: int
        Number of equidistant steps.
```

cuqi/geometry/_geometry.py
```diff
@@ -703,31 +703,49 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
+
+    axis_labels: list of strings
+       plotting axis labels
```
nabriis, line 716: I dont think we need to expose this? We can just use to default from Continuous1D right?

cuqi/geometry/_geometry.py
```diff
@@ -703,31 +703,49 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
+
+    kwargs: keyword arguments
+        keyword arguments defined in :class:`~cuqi.geometry.Continuous1D`
```
nabriis, line 719:
```suggestion
        keyword arguments are passed to the initializer of :class:`~cuqi.geometry.Continuous1D`
```

cuqi/geometry/_geometry.py
```diff
@@ -703,31 +703,49 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
         L = self.grid[-1]
-        self._idx1 = np.where( (self.grid>=0*L)&(self.grid<=0.333*L) )
-        self._idx2 = np.where( (self.grid>0.333*L)&(self.grid<=0.666*L) )
-        self._idx3 = np.where( (self.grid>0.666*L)&(self.grid<=L) )
+
+        self._ideces = []
+        for i in range(self._n_steps):
+            start = i*L/self._n_steps
+            end = (i+1)*L/self._n_steps
+            self._ideces.append( np.where((self.grid>=start)&(self.grid<=end))[0] )
```
nabriis, line 734: "indices" ?

nabriis (changes requested):
Hi @amal-ghamdi. Thanks for expanding the StepExpansion class. The PR looks good, I only had a few comments and suggested changes.

cuqi/geometry/_geometry.py
amal-ghamdi: Good point! I removed it.

cuqi/geometry/_geometry.py
amal-ghamdi: I just updated that :).

amal-ghamdi:
Thank you @nabriis for your helpful review! I added your suggestions!

nabriis (approved):

nabriis:
LGTM!

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,47 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
 
 class StepExpansion(Continuous1D):
     '''
-    class representation of the step random field with 3 intermidiate steps
+    Class representation of the step random field with n_steps intermediate steps.
+
+    Parameters:
+    -----------
+    grid: ndarray
+        Grid points for the step expansion to be evaluated at. There can be more grid points than number of steps.
```
jakobsj, line 710: The documentation would benefit from being expanded with: 1)  An conceptual explanation of what the "step random field" is. 2) Define what a "step" is and what is the "grid" and "grid point" is. 3) What is meant by "There can be more grid points than the number of steps" including why the user can specify both "grid" and "n_steps" as input (I would have expected either but not both, and I would have expected by specifying a grid that would be non-equidistant steps).

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,47 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
         L = self.grid[-1]
-        self._idx1 = np.where( (self.grid>=0*L)&(self.grid<=0.333*L) )
-        self._idx2 = np.where( (self.grid>0.333*L)&(self.grid<=0.666*L) )
-        self._idx3 = np.where( (self.grid>0.666*L)&(self.grid<=L) )
+
+        self._indices = []
+        for i in range(self._n_steps):
+            start = i*L/self._n_steps
+            end = (i+1)*L/self._n_steps
+            self._indices.append( np.where((self.grid>=start)&(self.grid<=end))[0] )
```
jakobsj, line 731: I would think only one of the lower and upper thresholds should include equality to avoid the risk of having a grid point assigned to mulitple intervals?

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,47 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
  
         return real
 
-    def fun2par(self,f):       
-        val1 = f[self._idx1[0][0]]
-        val2 = f[self._idx2[0][0]]
-        val3 = f[self._idx3[0][0]]
-        return np.array([val1,val2,val3])
+    def fun2par(self,f):
+        val = np.zeros(self._n_steps)
```
jakobsj, line 741: I think I may not be able to review the functionality until the documentation has been updated to clarify what grid points, steps etc are. I would like to return to the review then.

jakobsj (changes requested):
Thanks Amal, this is a very neat addition/generalization.  I tried to read the documentation pretending to be a completely new user, which prompted some questions. I'd like to take a look again once they have been addressed, so I can understand the new functionality better.

cuqi/geometry/_geometry.py
amal-ghamdi: Thank you @jakobsj for your review and the very helpful comments! I updated the documentation to give more explanation of the grid, steps, etc.

cuqi/geometry/_geometry.py
amal-ghamdi: Good point! I update that.

cuqi/geometry/_geometry.py
amal-ghamdi: Sounds great. I updated the documentation and it will be great if you can take a look and let me know what you think.

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,60 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
+    kwargs: keyword arguments
+        keyword arguments are passed to the initializer of :class:`~cuqi.geometry.Continuous1D`
     '''
-    def __init__(self, grid, axis_labels=['x'], **kwargs):
+    def __init__(self, grid, n_steps=3, **kwargs):
 
-        super().__init__(grid, axis_labels,**kwargs)
+        super().__init__(grid, **kwargs)
+        if n_steps > np.size(self.grid):
+            raise ValueError("n_steps must be smaller than the number of grid points")
```
jakobsj, line 732: Is this necessary? I suppose we could think of a possibly irregular grid, which happened to not have any grid points within one or more step intervals?

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,60 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
 
 class StepExpansion(Continuous1D):
     '''
-    class representation of the step random field with 3 intermidiate steps
+    Class representation of step functions (piecewise constant functions) with `n_steps` 
+    equidistant steps on the interval [0, L].
+    The function `par2fun` maps the parameters `p` (which are the step magnitudes) to the 
+    corresponding step function evaluated on the spacial grid (`grid`) of nodes x0=0, x1, ...xn=L.
```
jakobsj, line 708:
```suggestion
    corresponding step function evaluated on the spatial grid (`grid`) of nodes x0=0, x1, ...xn=L.
```

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,60 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
+    
+    For example, if `n_steps` is 3 and `grid` is a uniform grid with nodes x0=0, x1=0.1L, ..., xn=L, 
+    then the resulting function evaluated on the grid will be p[0], if x<=L/3, p[1], if L/3<x<=2L/3, 
+    p[2], if 2L/3<x<=L.
```
jakobsj, line 712: Nicely explained, thanks!

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,60 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
+    
+    Parameters:
+    -----------
+    grid: ndarray
+        Grid points for the step expansion to be evaluated at. There can be more grid points
```
jakobsj, line 717: There "can" be.. Must there be? The error raised below seems to indicate that the number of grid point MUST be larger than the number of steps.

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,60 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
+    then the resulting function evaluated on the grid will be p[0], if x<=L/3, p[1], if L/3<x<=2L/3, 
+    p[2], if 2L/3<x<=L.
+    
+    Parameters:
+    -----------
+    grid: ndarray
+        Grid points for the step expansion to be evaluated at. There can be more grid points 
+        than number of steps, which can be useful, for example, when using the StepExpansion 
+        geometry as a domain geometry for a cuqipy :class:`Model` that expects the input to be
+        interpolated on a (possibly fine) grid (`grid`).
```
jakobsj, line 720: Would it make sense to add that both interval endpoints must be included in `grid`?

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,60 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
 
-    def fun2par(self,f):       
-        val1 = f[self._idx1[0][0]]
-        val2 = f[self._idx2[0][0]]
-        val3 = f[self._idx3[0][0]]
-        return np.array([val1,val2,val3])
+    def fun2par(self,f):
+        val = np.zeros(self._n_steps)
+        for i in range(self._n_steps):
+            val[i] = f[self._indices[i][0]]
```
jakobsj, line 756: I wonder if taking the first element of the list is really the correct/best way of converting from fun to par. Seems to assume we have a function vector that is already constant on each step, which may not be the case. Should one perhaps instead average over the values within the step? This might be seen as some kind of projection on to the step expansion basis functions.

jakobsj (commented):
Thanks @amal-ghamdi, nice changes. I had time to take a very quick look and added some comments, let us chat later today.

cuqi/geometry/_geometry.py
amal-ghamdi: For now I made the setup such that the grid is regular (I added checks to ensure that). And for the regular grid case, possibly defining more steps than grid point is unnecessary. 

I am thinking possibly in the future we can generalize the code to accept irregular grid (including updating the projection to account for dx) and then we can look into this issue.

cuqi/geometry/_geometry.py
amal-ghamdi: I updated this documentation to clarify that the number of grid points can be equal n_steps to or more.

cuqi/geometry/_geometry.py
amal-ghamdi: Yes! I added that.

cuqi/geometry/_geometry.py
amal-ghamdi: This is a very good point! I kept it like this at first to keep the logic unchanged, but now I updated this to support mean, max, min, projection types.

amal-ghamdi:
Thank you @jakobsj for your helpful feedback! I made updates to mainly allow different projection (mean, max, min), and added documentation. I also added tests for the projection. Additionally, the interval now can start from point x0, and not necessary from 0. I appreciate if you can take a look.

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,80 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
 
 class StepExpansion(Continuous1D):
     '''
-    class representation of the step random field with 3 intermidiate steps
-    '''
-    def __init__(self, grid, axis_labels=['x'], **kwargs):
+    Class representation of step functions (piecewise constant functions) with `n_steps` 
+    equidistant steps on the interval [x0, xn], of length L.
```
jakobsj, line 706:
```suggestion
    equidistant steps on the interval [x0, xn], with both endpoints included.
```

jakobsj: I think it is no longer relevant to the user whether we refer to the interval length as `L` now that it is automatically computed from x0 and xn?

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,80 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
 
-        super().__init__(grid, axis_labels,**kwargs)
+    n_steps: int
+        | Number of equidistant steps.
 
-        L = self.grid[-1]
-        self._idx1 = np.where( (self.grid>=0*L)&(self.grid<=0.333*L) )
-        self._idx2 = np.where( (self.grid>0.333*L)&(self.grid<=0.666*L) )
-        self._idx3 = np.where( (self.grid>0.666*L)&(self.grid<=L) )
+    projection: str
```
jakobsj, line 722: Perhaps clearer if we called this e.g. `fun2par_projection`?

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,80 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
+        | Projection of the step function (evaluated on the grid) on the parameter space. The supported projections are 
+        | 'mean': the parameter p[i] value will be the average of the function values at the nodes that falls in the interval  I=(x0+i*L/n_steps, x0+(i+1)*L/n_steps].
```
jakobsj, line 724: Please state that mean is the default.

cuqi/geometry/_geometry.py
```diff
@@ -702,32 +702,80 @@ def _compute_eigpairs(self, xnod, C_nu, sigma, M, N_GL):
...
+        self._projection = projection
+        L = self.grid[-1]-self.grid[0]
+        x0 = self.grid[0]
+
+        self._indices = []
+        for i in range(self._n_steps):
+            start = x0 + i*L/self._n_steps
+            end = x0 + (i+1)*L/self._n_steps
+            if i ==0:
+                self._indices.append( np.where((self.grid>=start)&(self.grid<=end))[0] )
```
jakobsj, line 745: Bit hard to read this, especially the `[0]` part and understanding that `_indices` will be a list of length `n_steps`  with an array of indices in each element of `_indices. Can you please add a little bit of explanation  of this.

jakobsj (changes requested):
Thank you @amal-ghamdi , getting very close now. I think this should be captured by a separate issue: it is a bit odd to be allowing the user to pass a grid, which would seem to indicate a general grid can be given, then restricting it to a regular grid. In fact, I don't immediately see in the code that irregular grids will not work. But let us not leave it for another issue (please create) so we can get this closed with just the minor changes requested.

cuqi/geometry/_geometry.py
amal-ghamdi: great just did that

cuqi/geometry/_geometry.py
amal-ghamdi: I rewrote this part for readability and add a comment. The index [0] is just because in 1D array np.where returns for example (np.ndarray([1,3]),) where the first item in the tuple is the indices

amal-ghamdi:
Thank you @jakobsj for all your comments and feedback! I incorporated the last patch of your comments and suggestions.

cuqi/geometry/_geometry.py
```diff
@@ -728,23 +728,25 @@ class StepExpansion(Continuous1D):
...
-        self._projection = projection
+        self._fun2par_projection = fun2par_projection
         L = self.grid[-1]-self.grid[0]
         x0 = self.grid[0]
 
         self._indices = []
         for i in range(self._n_steps):
             start = x0 + i*L/self._n_steps
             end = x0 + (i+1)*L/self._n_steps
+            # Extract indices of the gird points that fall in the ith interval.
```
jakobsj, line 744:
```suggestion
            # Extract indices of the grid points that fall in the ith interval.
```

jakobsj (approved):
THank you, looks great! Just a typo then please merge.

amal-ghamdi:
Great! Thanks @jakobsj for your review. I created an issue #30 for supporting irregular grids (taking into account grid spacing when computing the mean projection).

PR #17, opened 2022-08-19
Automatically detect name of density

### LINKED ISSUE #20: Automatically determine variable name
Currently the user has the write the following code to define a random variable following a distribution e,g, `x ~ N(0,1)` 

``` Python
x = GaussianCov(0, 1, name="x") 
```

It would be good if the user could write
``` Python
x = GaussianCov(0, 1)
```
and the name is automatically determined.

nabriis:
Closes #20 

Currently the user has the write the following code to define a random variable following a distribution e,g, `x ~ N(0,1)` 

``` Python
x = GaussianCov(0, 1, name="x") 
```
With this update the user can write
``` Python
x = GaussianCov(0, 1)
```
and the name is automatically determined.

It required a bit of optimization and restructuring of the _condition method for distributions to make it work cleanly without too much overhead. After further discussion we also made the options for evaluating the logd and conditioning with positional and kwargs more consistent.

- [x] Added method for automatically determining python variable name of a density
- [x] Added tests
- [x] Added optimizations on conditioning for distributions to make it run faster (no functional change) 
- [x] Added functionality for evaluating with positions or kwargs

### DIFF
--- a/cuqi/density/_density.py
+++ b/cuqi/density/_density.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 from abc import ABC, abstractmethod
 from typing import Optional
+import cuqi
 
 class Density(ABC):
     """ Abstract base class for densities.
--- a/cuqi/density/_density.py
+++ b/cuqi/density/_density.py
@@ -106,7 +106,12 @@ class Density(ABC):
         return new_density
 
     def __call__(self, *args, **kwargs):
-        """ Condition on the given parameters. """
+        """ Condition the density on a set of parameters.
+        
+        Positional arguments must follow the order of the parameter names.
+        These can be accessed via the :meth:`get_parameter_names` method.
+        
+        """
         return self._condition(*args, **kwargs)
 
 class EvaluatedDensity(Density):
--- a/cuqi/density/_density.py
+++ b/cuqi/density/_density.py
@@ -39,7 +39,7 @@ class Density(ABC):
         if self._is_copy: # Extract the original density name
             return self._original_density.name
         if self._name is None: # If none extract the name from the stack
-            self._name = cuqi.utilities.get_python_variable_name(self)
+            self._name = cuqi.utilities._get_python_variable_name(self)
         return self._name
 
     @name.setter
--- a/cuqi/density/_density.py
+++ b/cuqi/density/_density.py
@@ -1,7 +1,9 @@
 from __future__ import annotations
 from abc import ABC, abstractmethod
 from typing import Optional
+from copy import copy
 import cuqi
+import ctypes
 
 class Density(ABC):
     """ Abstract base class for densities.
@@ -30,18 +32,27 @@ class Density(ABC):
             raise ValueError(f"{self.__init__.__qualname__}: Name must be a string or None")
         self.name = name
         self._constant = 0 # Precomputed constant to add to the log probability.
+        self._original_density = None # Original density if this is a conditioned copy.
 
     @property
     def name(self):
         """ Name of the random variable associated with the density. """
+        if self._is_copy: # Extract the original density name
+            return self._original_density.name
         if self._name is None: # If none extract the name from the stack
-            name = cuqi.utilities.get_python_variable_name(self)
-            self._name = name
+            self._name = cuqi.utilities.get_python_variable_name(self)
         return self._name
 
     @name.setter
     def name(self, name):
+        if self._is_copy:
+            raise ValueError("Cannot set name of conditioned density. Only the original density can have its name set.")
         self._name = name
+
+    @property
+    def _is_copy(self):
+        """ Returns True if this is a copy of another density. """
+        return hasattr(self, '_original_density') and self._original_density is not None
 
     def logd(self, *args, **kwargs):
         """ Evaluates the un-normalized log density function given a set of parameters.
@@ -86,6 +97,13 @@ class Density(ABC):
     def get_parameter_names(self):
         """ Returns the names of the parameters that the density can be evaluated at or conditioned on. """
         pass
+
+    def _make_copy(self):
+        """ Returns a shallow copy of the density keeping a pointer to the original. """
+        # Store id of original density
+        new_density = copy(self)
+        new_density._original_density = self
+        return new_density
 
     def __call__(self, *args, **kwargs):
         """ Condition on the given parameters. """
--- a/cuqi/density/_density.py
+++ b/cuqi/density/_density.py
@@ -3,7 +3,6 @@ from abc import ABC, abstractmethod
 from typing import Optional
 from copy import copy
 import cuqi
-import ctypes
 
 class Density(ABC):
     """ Abstract base class for densities.
--- a/cuqi/density/_density.py
+++ b/cuqi/density/_density.py
@@ -31,14 +31,14 @@ class Density(ABC):
             raise ValueError(f"{self.__init__.__qualname__}: Name must be a string or None")
         self.name = name
         self._constant = 0 # Precomputed constant to add to the log probability.
-        self._original_density = None # Original density if this is a conditioned copy.
+        self._original_density = None # Original density if this is a conditioned copy. Used to extract name.
 
     @property
     def name(self):
         """ Name of the random variable associated with the density. """
         if self._is_copy: # Extract the original density name
             return self._original_density.name
-        if self._name is None: # If none extract the name from the stack
+        if self._name is None: # If None extract the name from the stack
             self._name = cuqi.utilities._get_python_variable_name(self)
         return self._name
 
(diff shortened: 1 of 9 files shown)


nabriis:
Hi @amal-ghamdi. I added you as reviewer on this PR. Its pointing to the other PR until it gets merged into main.

cuqi/distribution/_distribution.py
```diff
@@ -267,11 +266,32 @@ def _condition(self, *args, **kwargs):
...
-        # If conditioning on the name of the distribution
-        # we convert the distribution to a likelihood.
-        if self.name in kwargs:
-            return new_dist.to_likelihood(kwargs[self.name])               
+                
+                # Store processed keywords
+                processed_kwargs.update(var_args.keys())
+
+        # Check if more arguments were passed than conditioning variables
+        # If so we assume the last one is the distribution name
```
amal-ghamdi, line 274: This looks like a reasonable convention. I wanted to ask if it is it documented for the user that the last argument is the calling distribution random variable realization (name).

cuqi/utilities/_get_python_variable_name.py
```diff
@@ -0,0 +1,45 @@
+import sys
+from itertools import count
+import re
+
+import warnings
+import cuqi
+
+
+def get_python_variable_name(var):
```
amal-ghamdi, line 9: I suggest to make this hidden method '_get_python_variable_name', since we do not expect users to use it.

cuqi/utilities/_get_python_variable_name.py
```diff
@@ -0,0 +1,45 @@
...
+            
+            if len(var_names) > 0:
+                return var_names[0]
+
+    warnings.warn("Could not automatically find variable name for object: {}. If code runs slowly and variable name is not needed set config.MAX_STACK_SEARCH_DEPTH to 0.".format(var))
+
+    return None
+
+#https://stackoverflow.com/questions/34115298/how-do-i-get-the-current-depth-of-the-python-interpreter-stack
+def stack_size2a(size=2):
```
amal-ghamdi, line 37: I suggest to make this hidden method '_stack_size2a', since we do not expect users to use it.

cuqi/utilities/_get_python_variable_name.py
```diff
@@ -0,0 +1,45 @@
...
+    """Get stack size for caller's frame.
+    """
+    frame = sys._getframe(size)
```
amal-ghamdi, line 40: Is `size` here represent max size we are searching for. If so, may be rename it `max_size`? if that makes sense.

tests/test_abstract_distribution_density.py
```diff
@@ -32,10 +32,21 @@ def test_conditioning_wrong_keyword():
         x(name="hey") #Should expect value-error since name not conditioning variable.
 
 def test_conditioning_arg_as_mutable_var():
-    """ This checks if we raise error if arg is given for a distribution that has no conditioning variables. """
+    """ This checks if we raise error if 2 args is given for a distribution that has no conditioning variables. """
```
amal-ghamdi, line 35:
```suggestion
    """ This checks if we raise error if 2 args are given for a distribution that has no conditioning variables. """
```

cuqi/distribution/_posterior.py
```diff
@@ -78,9 +78,15 @@ def geometry(self, value):
         else:
             self._geometry = self.prior.geometry
             
-    def logpdf(self, x):
+    def logpdf(self, *args, **kwargs):
         """ Returns the logpdf of the posterior distribution"""
-        return self.likelihood.logd(x)+ self.prior.logd(x)
+        return self.likelihood.logd(*args, **kwargs)+ self.prior.logd(*args, **kwargs)
```
amal-ghamdi, line 83: is it going to be problematic that we pass `*arg` and `**kwargs` of the posterior to both the prior and likelihood? is this going to cause passing irrelevant args and kwargs to the prior for example (and to the likelihood )

cuqi/distribution/_posterior.py
```diff
@@ -78,9 +78,15 @@ def geometry(self, value):
...
+
+    def get_conditioning_variables(self):
+        return self.prior.get_conditioning_variables()
+
+    def get_parameter_names(self):
```
amal-ghamdi, line 88: In these two methods, `get_conditioning_variables` and `get_parameter_names`

Is it possible that the likelihood has additional conditioning variables ?

cuqi/density/_density.py
```diff
@@ -30,6 +31,18 @@ def __init__(self, name: Optional[str] = None):
         self.name = name
         self._constant = 0 # Precomputed constant to add to the log probability.
 
+    @property
+    def name(self):
+        """ Name of the random variable associated with the density. """
+        if self._name is None: # If none extract the name from the stack
+            name = cuqi.utilities.get_python_variable_name(self)
```
amal-ghamdi, line 38: if the name is None after search, may be we we can print a warning message to suggest that the user can specify the distribution name if needed.

amal-ghamdi (commented):
Thank you @nabriis for adding this great feature! And it is great that you optimized it carefully.

I had few comments that I wrote throughout and two questions


- If we search for the variable name first thing in the initializer (for example by calling self.name first thing in the initializer), Do you think that will reduce the the level needed to traverse the stack. Just a question that came across my mind.

- I looked a bit at the naming and for example when I run
```python
dist = cuqi.distribution.GaussianCov()
print(dist.name)
y = dist(mean=1)
print(dist.name)
print(y.name) 
```
y name will be 'dist'

but when I run same code but with out calling dist.name at the second line
```python
dist = cuqi.distribution.GaussianCov()
#print(dist.name)
y = dist(mean=1)
print(dist.name)
print(y.name)
```
y name will be 'y'. Should we make it consistent?

cuqi/distribution/_distribution.py
nabriis: It is now :)

cuqi/utilities/_get_python_variable_name.py
nabriis: Size is the size of the current stack. I think its OK to leave it as is

cuqi/distribution/_posterior.py
nabriis: This is a good point. I will add a check during the initializer that ensures the likelihood parameters match a distribution.

cuqi/density/_density.py
nabriis: Good suggestion. There is a warning message in the python variable name function. I updated it a bit.

cuqi/distribution/_posterior.py
nabriis: For now this is OK because the likelihood and prior have the same arguments. In the long term Posterior will likely be replaced completely by joint distribution I would imagine.

nabriis:
> Thank you @nabriis for adding this great feature! And it is great that you optimized it carefully.
> 
> I had few comments that I wrote throughout and two questions
> 
> * If we search for the variable name first thing in the initializer (for example by calling self.name first thing in the initializer), Do you think that will reduce the the level needed to traverse the stack. Just a question that came across my mind.
> * I looked a bit at the naming and for example when I run
> 
> ```python
> dist = cuqi.distribution.GaussianCov()
> print(dist.name)
> y = dist(mean=1)
> print(dist.name)
> print(y.name) 
> ```
> 
> y name will be 'dist'
> 
> but when I run same code but with out calling dist.name at the second line
> 
> ```python
> dist = cuqi.distribution.GaussianCov()
> #print(dist.name)
> y = dist(mean=1)
> print(dist.name)
> print(y.name)
> ```
> 
> y name will be 'y'. Should we make it consistent?

Hi @amal-ghamdi. Thanks for the review.

First of all, is is *not* possible to extract the name while the class instance is being created. The reason for this is that the variable `x` when running `x=GaussianCov()` is not set until after initilization.

Second, thanks for catching this inconsistency in the name determination. I updated the way to determine name so that it is always the first definition of a distribution that defines its name.

nabriis:
Hi @jakobsj. I added you as second reviewer. There should be many comments in the code, but there is still the main question if Likelihood and EvaluatedDensity should even have a name attribute. 

Currently for Likelihood the name attribute is set to be the name of the underlying distribution. If name strictly means the "RandomVariable" I think this does not make sense since there is no random variable associated with a Likelihood (afaik). However, in the current upcoming code for Joint Distribution we do use name to keep track of each of the densities in the "Bayesian model". For this reason I would likely suggest deferring this to an issue and handling it later. If you think of a good suggestion though, let me know :)

jakobsj (approved):
I tried it out and it appears to work well. Went through the code, which is somewhat difficult to come into and follow but I didn't see any problems, only mentioned a typo. Re. likelihood and evaluated density, currently at least for evaluated density I think it makes sense that the name of the random variable is still stored, probably also for likelihood. There is a question as to whether "name" is the right term to use here, or it might be "rv" or "rvname" etc. But do go ahead, great work.

amal-ghamdi (approved):
Thank you @nabriis for the update and the clarification. I think keeping track of the original density is a good idea. The PR LGTM!

PR #23, opened 2022-08-24
Sprint10 plot ci func new

### LINKED ISSUE #10: Add option to compute ci for function field

amal-ghamdi:
closes #10 

Main updates:

- [x] Add `funvals` and `is_par` to the samples object 
- [x] Allow computing statistics on `samples.funvals` as long as the function space of the geometry is represented by a one dimensional ndarray. 
- [x] Distinguish fun_dim and par_dim, and fun_shape and par_shape for `geometry` and `likelihood`.
    - [x] `par_shape` need to be defined, `par_dim` is derived from `par_shape`. `fun_shape` is inferred if possible or defines, `fun_dim` is inferred from `fun_shape`.
    - [x] Update demos, tests and other modules that uses dim/shape of geometry to implement the correct use.


amal-ghamdi:
Hi @jakobsj and @nabriis, based on our discussion on computing statistics on function values, this MR has the implementation that achieves this via creating a new samples object (you can see the try script for example.). I think the code here is much simpler ( compared to PR #11 I removed getter and setter of samples property, removed the flag `compute_on_par`, removed updates for `compute_ci` and `mean`, removed updates in burnthin because all those updates are not relevant now). I closed PR #11 and I appreciate your review on this new one.

cuqi/geometry/_geometry.py
```diff
@@ -34,6 +35,11 @@ def variables(self):
                 self.variables = self.dim
         return self._variables
 
+    @property
+    def fun_dim(self):
+        """The dimension of the function space."""
+        return self.dim
```
nabriis, line 41: I am wary of assuming the default function value dimension is equal to the parameters. Perhaps it would be better to infer this property from `par2fun` by default instead? 

One suggestion would be something like
``` Python
    @property
    def fun_shape(self):
        """ The shape of the geometry (function space). """
        if self._fun_shape is None:
            # Attempt to infer dimension
            try:
                self._fun_shape = self.par2fun(np.ones(self.dim)).shape
            except Exception:
                pass
        return self._fun_shape
    
    @property
    def fun_dim(self):
        """ The dimension of the geometry (function space). """
        if self.fun_shape is None: return None
        return reduce(operator.mul, self.fun_shape) # math.prod(self.shape) for Python 3.8+
```


In fact we need to think about the parameter and function space shape and dimensions. I think these are inconsistent in various geometries matching either the parameter space or function space. It would be very very good to fix this. For example, currently the main assumption is that the `shape` attribute represents the *function space* shape in Image2D. In just turns out that `dim` matches the parameter space for this case, but that may not always be the case.

I would suggest we rethink this as part of this PR. In particular we can use names like `shape_par` and `shape_fun`, and `dim_par` and `dim_fun` to match. Let is discuss ASAP

cuqi/samples/_samples.py
```diff
@@ -190,6 +194,24 @@ def geometry(self):
             self._geometry = _DefaultGeometry(grid=np.prod(self.samples.shape[:-1]))
         return self._geometry
 
+    @property
+    def funvals(self):
```
nabriis, line 198: We should add a similar method for `.parameters` matching the implementation in CUQIArray.

cuqi/samples/_samples.py
```diff
@@ -190,6 +194,24 @@ def geometry(self):
...
+            return Samples(_funvals, is_par=False, geometry=self.geometry)
+        else:
+            return self
+
+    @property
+    def geometry_dim(self):
+        if self.is_par:
+            return self.geometry.dim
+        else:
+            return self.geometry.fun_dim
```
nabriis, line 213: Do we really need this method? At least I do not think it should be public. If it is used in various places we can make it private. Also is it really the `dim` we are looking for or the `shape`?

cuqi/samples/_samples.py
```diff
@@ -325,7 +347,7 @@ def plot(self,sample_indices=None,*args,**kwargs):
 
 
     def plot_chain(self, variable_indices=None, *args, **kwargs):
-        dim = self.geometry.dim
```
nabriis, line 328: I think this will cause issues for function value Samples objects that are multi dimensional as is the case for e.g. `Image2D`. Perhaps plot_chain should not work for funvals samples arrays.

cuqi/samples/_samples.py
```diff
@@ -370,12 +390,35 @@ def plot_ci(self,percent=95,exact=None,*args,plot_envelope_kwargs={},**kwargs):
...
+
+        # is_par is determined automatically from self.is_par 
+        if "is_par" in kwargs.keys() or\
+           "is_par" in pe_kwargs.keys():
+            raise ValueError("The flag `is_par` is determined automatically and should not be passed to `plot_ci`.")
+
+        #User cannot ask for computing statistics on function values then plotting on parameter space
+        if not self.is_par:
+            if "plot_par" in kwargs.keys() and kwargs["plot_par"] or\
+                    "plot_par" in pe_kwargs.keys() and kwargs["plot_par"]:
```
nabriis, line 405:
```suggestion
            if "plot_par" in kwargs and kwargs["plot_par"] or\
                    "plot_par" in pe_kwargs and kwargs["plot_par"]:
```

cuqi/samples/_samples.py
```diff
@@ -370,12 +390,35 @@ def plot_ci(self,percent=95,exact=None,*args,plot_envelope_kwargs={},**kwargs):
...
+                    "Cannot plot credible interval on parameter space if the samples are in the function space.")
+
+        # Depending on the value of self.is_par, the computed statistics below (mean, lo_conf,up_conf) are either parameter
+        # values or function values
+        statistics_is_par = self.is_par
+        pe_kwargs["is_par"] = statistics_is_par
+        kwargs["is_par"] = statistics_is_par
+
+        # Set plot_par value to be passed to Geometry.plot_envelope and Geometry.plot.
+        if "plot_par" in kwargs.keys():
```
nabriis, line 417:
```suggestion
        if "plot_par" in kwargs:
```

cuqi/samples/_samples.py
```diff
@@ -399,17 +442,19 @@ def plot_ci(self,percent=95,exact=None,*args,plot_envelope_kwargs={},**kwargs):
...
+            lci = self.geometry.plot_envelope(
+                lo_conf, up_conf, color='dodgerblue', **pe_kwargs)
             
-            lmn = self.geometry.plot(mean,*args,**kwargs)
-            if exact is not None: #TODO: Allow exact to be defined in different space than mean?
+            lmn = self.geometry.plot(mean, *args, **kwargs)
+            if exact is not None:  #TODO: Allow exact to be defined in different space than mean?
                 if isinstance(exact, CUQIarray):
-                    lex = exact.plot(*args,**kwargs)
+                    lex = exact.plot(*args, **kwargs)
```
nabriis, line 451: Should this check for plot_par / is_par? We would want the exact to be expressed on the same space

cuqi/samples/_samples.py
```diff
@@ -438,7 +483,7 @@ def plot_autocorrelation(self, variable_indices=None, max_lag=None, combined=Tru
         -------
         axes: matplotlib axes or bokeh figures
         """
-        dim = self.geometry.dim
+        dim = self.geometry_dim
```
nabriis, line 486: Again. Does it make sense to plot these things for funvals? And will it work for n-d arrays?

nabriis (commented):
Hi @amal-ghamdi,

Thank you very much for this addition to the Samples class. Going through these changes made me realize that we have some inconsistencies that in the `Geometry` class from way back.

I think it would be worthwhile to deal with these while we are at it with this PR. For this reason I added quite a number of comments. Do not fret! We can discuss and work on it together.

cuqi/geometry/_geometry.py
amal-ghamdi: Thank you @nabriis for your suggestion. This is a very good point! Yes I think it will be nice to address it. Lets discuss this sometime soon.

amal-ghamdi: This is great point, now we have 4 properties (fun_shape, fun_dim, par_shape, par_dim). `par_shape` need to be defined, `par_dim` is derived from `par_shape`. `fun_shape` is inferred if possible or defines, `fun_dim` is inferred from `fun_shape`.

Now calling `funvals` and setting `Samples.is_par` to False, is only allowed for geometries where the function space is represented by a 1D ndarray.

cuqi/samples/_samples.py
amal-ghamdi: good point! I added this functionality.

cuqi/samples/_samples.py
amal-ghamdi: I made it hidden for now.

cuqi/samples/_samples.py
amal-ghamdi: Great catch! Now `funvals` only works for geometries with functions represented by 1D ndarray

cuqi/samples/_samples.py
amal-ghamdi: There are checks for is_par and plot_par and setting them automatically if needed in this method (let me know if this does not solve the problem you are thinking about). One thing I was thinking about is removing the args from `lex = exact.plot(*args, **kwargs)` and from plot_ci signature. I think is is confusion since we have two plotting types here, plot envelope and the regular plot.

cuqi/samples/_samples.py
amal-ghamdi: for now plotting on function space is restricted to geometries where the function space is represented by a 1D ndarray.

amal-ghamdi:
Thank you @nabriis for your thorough review and discussion about this PR!. I addressed the points you mentioned and I greatly appreciate it if you can provide a re-review.

cuqi/geometry/_geometry.py
```diff
@@ -256,6 +278,12 @@ def _create_dimension(self, dim_grid):
     @property
     def grid(self):
         return self._grid
+
+    @property
+    def par_shape(self):
+        """The shape of the parameter space."""
+        if self.fun_dim is None: return None  
+        return (self.fun_dim,)
```
nabriis, line 286: This confuses me. Why are we using fun_dim to define the parameter dimension? To me it would make more sense to define par_shape as the original parameter and never infer that from anything else. This way the code is more explicit rather than implicit.

I guess it might mean you have to extract the parameters differently e.g:

``` Python
    @property
    def par_shape(self):
        """The shape of the parameter space."""
        if self.grid is None: return None
        return (np.size(self.grid), )
```

and

``` Python
    @property
    def par_shape(self):
        """The shape of the parameter space."""
        if self.grid is None: return None
        return np.shape(self.grid)
```

Perhaps the above approach works ok, but it is rather hard to figure out what is going on, which means at a later point we may run into errors we do not understand

cuqi/geometry/_geometry.py
```diff
@@ -387,29 +415,33 @@ class Image2D(Geometry):
...
-    @shape.setter
-    def shape(self, value):
-        self._shape = value
+    @fun_shape.setter
+    def fun_shape(self, value):
+        self._fun_shape = value
+
+    @property
+    def par_shape(self):
+        return (self.fun_dim,)
```
nabriis, line 440: Same comment as above

cuqi/geometry/_geometry.py
```diff
@@ -442,9 +474,16 @@ def __init__(self,variables):
...
+    def fun_shape(self):
+        """The shape of the function space."""
         return (len(self.variables),)
 
+    @property
+    def par_shape(self):
+        """The shape of the parameter space."""
+        if self.fun_dim is None: return None  
+        return (self.fun_dim,) 
+
```
nabriis, line 486: Same comment as above

cuqi/samples/_samples.py
nabriis: Got it. I missed adding the plot_par to kwargs

nabriis (commented):
Hi @amal-ghamdi. I went through the code and I think its nearly there. Two suggestions:

1) Add test for the following equality between samples.funvals.parameters and samples (when imap is defined in geometry).
2) Add some documentation to Geometry base class about par_shape and fun_shape.

cuqi/geometry/_geometry.py
nabriis: Great!

cuqi/geometry/_geometry.py
amal-ghamdi: This is a great point. Now par_shape implementation  is defined independent of `fun_dim`

amal-ghamdi:
Thank you @nabriis for your review! The point about making par_shape independent from fun_dim is a very good point! I also removed the implementation of par_shape from Continuous because there is no need for it there. It is implemented in the subclasses `Continuous1D` and `Continuous2D`. 

I added the documentation and the test as you suggested.

I appreciate if you can take a look.

nabriis (approved):
Hi @amal-ghamdi,

Thanks for adding this important update to the sampler class. After the proposed changes the PR now LGTM!

amal-ghamdi:
Thank you @nabriis for your review and very helpful comments!

demos/try16_Samples_funvals_stat.py
```diff
@@ -0,0 +1,26 @@
...
+sys.path.append("..")
+import cuqi
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+samples = cuqi.samples.Samples(np.random.rand(10, 15), geometry = cuqi.geometry.MappedGeometry(cuqi.geometry.Continuous1D(10), map=lambda x:x**2))
+
+# Compute statistics on parameter values
+samples.plot_ci(95, plot_par = True) #plot in parameter space
```
jakobsj, line 13: If the geometry is changed to
```
samples = cuqi.samples.Samples(3*np.random.rand(imsize**2, 15), geometry = cuqi.geometry.Continuous2D(grid=(4,4)))
```
we get
image
which appears to use parameters for the mean but functional values for the credibility interval. I know we discussed this PR would only deal with 1D-like geometry and I am unsure whether Continuous2D is included, but it seems unintuitive that `plot_par` only applies to the mean part. I know that `.funvals` does not yet work for 2D geometries, but 
```
samples.plot_ci(95)
```
works as I would expect giving all function value plots, so I wonder if the `plot_par` version can be made to plot all in parameter style in this PR?

jakobsj (changes requested):
Looking good, a few minor comments and one question about the case of plotting with plot_par=True for a Continuous2D geometry. Maybe I missed if this was now the expected behaviour or a fix postponed, so please just let me know if that is the case. Otherwise very close to mergeable.

cuqi/geometry/_geometry.py
```diff
@@ -7,6 +7,7 @@
 import scipy.sparse as sparse
 import operator
 from functools import reduce
+import warnings
 
 def _get_identity_geometries():
     """Return the geometries that have identity par2fun and fun2par methods (can still reshape).
```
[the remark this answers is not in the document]
amal-ghamdi: Good point! I updated the description

cuqi/geometry/_geometry.py
```diff
@@ -16,24 +17,48 @@ def _get_identity_geometries():
...
 class Geometry(ABC):
     """A class that represents the geometry of the range, domain, observation, or other sets.
+
+    It specifies a mapping from the parameter space to the function space (:meth:`par2fun`) and the inverse map if possible (:meth:`fun2par`). The parameters can be for example, the center and width of a hat function, and the function is the resulting hat function evaluated at grid points of a given grid. The geometry keeps track of the dimension and shape of the parameter space (:meth:`par_dim` and :meth:`par_shape`) and the dimension and shape of the function space (:meth:`fun_dim` and :meth:`fun_shape`).
     """
     @property
     @abstractmethod
-    def shape(self):
+    def par_shape(self):
+        """The shape of the parameter space."""
```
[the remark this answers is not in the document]
amal-ghamdi: I agree with you that `par_shape` does not seem to be needed now. It is added to provide the full for options. We can keep it just in case and might remove it in the future.

demos/try16_Samples_funvals_stat.py
amal-ghamdi: Thank you for investigating this! This is not an expected behavior (and it is related to the `plot_par` flag). This PR did not change this behavior (as this PR is concerned with computing statistics on function values) but I think it is better to fix it now rather than later. So I added one line change and it seems working is expected now: when  `plot_par` is True, no 2D images will appear.

amal-ghamdi:
Thank you @jakobsj for your very helpful feedback and investigating the plot_par behavior with 2D geometries! I incorporated your comments and I appreciate if you can take a look at the updates.

jakobsj (approved):
Thank you, looks good to go now!

amal-ghamdi:
Great, thank you @jakobsj for your feedback!

PR #38, opened 2022-08-31
Fix bug with gradient not returning to parameter space #37

### LINKED ISSUE #37: Model gradient returns output in wrong shape if working with 2D image geometry
This shows e.g. in the try12 script where the NUTS sampler fails for the 2D image problems.

I tracked it down to an issue with model.gradient, where if both range and domain geometry are image2D, given input in parameter space (N^2, ), the output will be (N,N), i.e. function space, instead of (N^2,) parameter space.

nabriis:
Closes #37


amal-ghamdi (approved):
nice catch :)

PR #45, opened 2022-09-01
Sprint11 Add repr to Posterior

### LINKED ISSUE #46: Add repr to Posterior
Current our Posterior distribution does not have any repr. It would be good to add a repr that matches the Joint Distribution (but only for a single parameter),

nabriis:
Closes #46.

Also removed extra docs that we automatically infer now in our online docs.


amal-ghamdi (approved):
LGTM!

PR #56, opened 2022-09-07
ensure grid type is float

### LINKED ISSUE #55: ensure grid type is float in continuous geometries and make parameters array type in StepExpansion always float

amal-ghamdi:
closes #55

### DIFF
--- a/cuqi/geometry/_geometry.py
+++ b/cuqi/geometry/_geometry.py
@@ -276,7 +276,7 @@ class Continuous(Geometry, ABC):
                 raise ValueError(dim_grid_value_err_msg)
         else:
             raise ValueError(dim_grid_value_err_msg)
-        return dim_grid
+        return dim_grid.astype(np.float64)
         
     @property
     def grid(self):
@@ -798,7 +798,7 @@ class StepExpansion(Continuous1D):
         return (self._n_steps,)
 
     def par2fun(self, p):
-        real = np.zeros_like(self.grid)  
+        real = np.zeros(self.grid.shape)
         for i in range(self._n_steps):
             real[self._indices[i]] = p[i]
  
--- a/cuqi/geometry/_geometry.py
+++ b/cuqi/geometry/_geometry.py
@@ -276,7 +276,7 @@ class Continuous(Geometry, ABC):
                 raise ValueError(dim_grid_value_err_msg)
         else:
             raise ValueError(dim_grid_value_err_msg)
-        return dim_grid.astype(np.float64)
+        return dim_grid.astype(float)
         
     @property
     def grid(self):


amal-ghamdi:
Thank you @nabriis for taking a look. 

**Fix**
The main fix in this PR is `real = np.zeros(self.grid.shape)`. It was before: `real = np.zeros_like(self.grid)`. 

before change, if self.grid is int, parameter values will be int. Which means if the function has value 1.5 in one of the steps, the resulting projection on the parameter space will make it 1 (float -> int).

**New convention**
Then I realized to ensure consistency, and to avoid any similar future problems we can make the convention that the grid values is always float. Thus the change: `return dim_grid.astype(float)` (note: I just change it from np.float64 to float)

nabriis:
> Thank you @nabriis for taking a look.
> 
> **Fix** The main fix in this PR is `real = np.zeros(self.grid.shape)`. It was before: `real = np.zeros_like(self.grid)`.
> 
> before change, if self.grid is int, parameter values will be int. Which means if the function has value 1.5 in one of the steps, the resulting projection on the parameter space will make it 1 (float -> int).
> 
> **New convention** Then I realized to ensure consistency, and to avoid any similar future problems we can make the convention that the grid values is always float. Thus the change: `return dim_grid.astype(float)` (note: I just change it from np.float64 to float)

Hi @amal-ghamdi, OK. It looks good to me. I wondered if it could cause some issues for e.g. large 2D geometries forcing everything to float, but its probably fine.

jakobsj (approved):
LGTM

amal-ghamdi:
> > Thank you @nabriis for taking a look.
> > **Fix** The main fix in this PR is `real = np.zeros(self.grid.shape)`. It was before: `real = np.zeros_like(self.grid)`.
> > before change, if self.grid is int, parameter values will be int. Which means if the function has value 1.5 in one of the steps, the resulting projection on the parameter space will make it 1 (float -> int).
> > **New convention** Then I realized to ensure consistency, and to avoid any similar future problems we can make the convention that the grid values is always float. Thus the change: `return dim_grid.astype(float)` (note: I just change it from np.float64 to float)
> 
> Hi @amal-ghamdi, OK. It looks good to me. I wondered if it could cause some issues for e.g. large 2D geometries forcing everything to float, but its probably fine.

Thank you @nabriis for your comment!

I think it is probably fine in terms of data size given that a single 2D sample will have l*m parameters and the grid information is stored in l+m so it is relatively small. And we store one geometry per n samples. 

Also since  grid points in continuous geometry represents points at R or R^2, it might makes sense to store them as floats.

nabriis (approved):

PR #59, opened 2022-09-07
Sprint11 add violin plot

### LINKED ISSUE #53: Add violin plot
Currently needed for notebook 2, but can be removed if not feasible to add by deadline.

nabriis:
Closes #53

### DIFF
--- a/cuqi/samples/_samples.py
+++ b/cuqi/samples/_samples.py
@@ -706,6 +706,40 @@ class Samples(object):
             RHAT[i] = value.to_numpy()
         return RHAT
 
+    def plot_violin(self, variable_indices=None, **kwargs):
+        """ Create a violin plot of the samples. 
+        
+        Parameters
+        ----------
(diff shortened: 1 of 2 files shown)


jakobsj (approved):
I tested in the notebook 2, and it works fine for both the 1D case so the current notebook runs. I also tested it for the 3-element Gaussian in that notebook, also works fine and makes a more interesting plot, so I'm thinking to update the notebook to do the violin plot for the 3-element instead of the 1D case.

nabriis:
> I tested in the notebook 2, and it works fine for both the 1D case so the current notebook runs. I also tested it for the 3-element Gaussian in that notebook, also works fine and makes a more interesting plot, so I'm thinking to update the notebook to do the violin plot for the 3-element instead of the 1D case.

Sounds good :) Wait until after my review on nb2 please.

amal-ghamdi (approved):
Thanks @nabriis, LGTM!

PR #62, opened 2022-09-07
update Heat_1D, Poisson_1D test problems documentation, parameters, __init__

### LINKED ISSUE #58: Add changes to testproblem Heat_1D required for notebook 5
- [x] Ability to pass `n_steps`
- [x] Ability to pass observation nodes map (instead of num_obs)
- [x] change parameters names: `KL_map` and `KL_imap` to `map` and `imap`
- [x] Update documentation
- [x] Update poisson_1D test problem to be consistent with `Heat_1D`
- [x] Update test

amal-ghamdi:
closes #58

### DIFF
--- a/cuqi/geometry/_geometry.py
+++ b/cuqi/geometry/_geometry.py
@@ -593,7 +593,7 @@ class KLExpansion(Continuous1D):
     .. math::
         f = \sum_{i=0}^{N-2} \\left(\\frac{1}{(i+1)^\\gamma\\tau}\\right)  p_i \\, \\text{sin}\\left(\\frac{\\pi}{N}(i+1)(K+\\frac{1}{2})\\right) 
         
-        + \\frac{(-1)^K}{2}(\\frac{1}{(N)^\\gamma\\tau})  p_{N-1}
+        + \\frac{(-1)^K}{2}\\left(\\frac{1}{N^\\gamma\\tau}\\right)  p_{N-1}
 
     where:
     :math:`K=\\{0, 1, 2, 3, ..., N-1\\}` and :math:`N` is the number of nodes in the grid and :math:`p_i` are the expansion coefficients. 
@@ -657,7 +657,7 @@ class KLExpansion_Full(Continuous1D):
     .. math::
         f = \\frac{\\text{std}^2}{\\pi}\sum_{i=0}^{N-2} \\left(\\frac{\\tau^\\gamma}{(\\tau+i^2)^\\gamma}\\right)  p_i \\, \\text{sin}\\left(\\frac{\\pi}{N}(i+1)(K+\\frac{1}{2})\\right) 
         
-        + \\frac{\\text{std}^2}{\\pi}\\frac{(-1)^K}{2}(\\frac{\\tau^\\gamma}{(\\tau+(N-1)^2)^\\gamma}) p_{N-1}
+        + \\frac{\\text{std}^2}{\\pi}\\frac{(-1)^K}{2}\\left(\\frac{\\tau^\\gamma}{\\left(\\tau+(N-1)^2\\right)^\\gamma}\\right) p_{N-1}
 
     where:
     :math:`K=\\{0, 1, 2, 3, ..., N-1\\}` and :math:`N` is the number of nodes in the grid and :math:`p_i` are the expansion coefficients. 
--- a/cuqi/geometry/_geometry.py
+++ b/cuqi/geometry/_geometry.py
@@ -595,8 +595,7 @@ class KLExpansion(Continuous1D):
         
         + \\frac{(-1)^K}{2}\\left(\\frac{1}{N^\\gamma\\tau}\\right)  p_{N-1}
 
-    where:
-    :math:`K=\\{0, 1, 2, 3, ..., N-1\\}` and :math:`N` is the number of nodes in the grid and :math:`p_i` are the expansion coefficients. 
+    where :math:`K=\\{0, 1, 2, 3, ..., N-1\\}` and :math:`N` is the number of nodes in the grid and :math:`p_i` are the expansion coefficients. 
 
     The above transformation is the inverse of DST-II (see https://en.wikipedia.org/wiki/Discrete_sine_transform)
 
@@ -659,8 +658,7 @@ class KLExpansion_Full(Continuous1D):
         
         + \\frac{\\text{std}^2}{\\pi}\\frac{(-1)^K}{2}\\left(\\frac{\\tau^\\gamma}{\\left(\\tau+(N-1)^2\\right)^\\gamma}\\right) p_{N-1}
 
-    where:
-    :math:`K=\\{0, 1, 2, 3, ..., N-1\\}` and :math:`N` is the number of nodes in the grid and :math:`p_i` are the expansion coefficients. 
+    where :math:`K=\\{0, 1, 2, 3, ..., N-1\\}` and :math:`N` is the number of nodes in the grid and :math:`p_i` are the expansion coefficients. 
 
     The above transformation is the inverse of DST-II (see https://en.wikipedia.org/wiki/Discrete_sine_transform)
 
--- a/cuqi/geometry/_geometry.py
+++ b/cuqi/geometry/_geometry.py
@@ -581,22 +581,52 @@ class _DefaultGeometry(Continuous1D):
 #         return self.mean + self.L.T@p
     
 class KLExpansion(Continuous1D):
-    '''
-    class representation of the random field in  the sine basis
-    alpha = sum_i p * (1/i)^decay * sin(i*L*x/pi)
-    L is the length of the interval
-    '''
+    """
+    Class representation of the random field in the sine basis
+
+    .. math::
+        f = \sum_{i=0}^{N-2} (\\frac{1}{(i+1)^\\text{decay}\\text{normalizer}})  p_i \\, sin(\\frac{\\pi}{N}(i+1)(K+\\frac{1}{2})) 
+        
+        + \\frac{(-1)^K}{2}(\\frac{1}{(N)^\\text{decay}\\text{normalizer}})  p_{N-1}
+
+    where:
+    :math:`K=\\{0, 1, 2, 3, ..., N-1\\}` and :math:`N` is the number of nodes in the grid and :math:`p_i` are the expansion coefficients. 
+
+    The above transformation is the inverse of DST-II (see https://en.wikipedia.org/wiki/Discrete_sine_transform)
+
+    Parameters:
+    -----------
+    grid : array-like
+        One dimensional regular grid on which the random field is defined.
+
+    decay_rate : float, default 2.5
+        The decay rate of the basis functions.
+
+    normalizer : float, default 1.0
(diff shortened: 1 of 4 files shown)


amal-ghamdi:
Hi @jakobsj and @nabriis. I appreciate it if you can review this PR. Issue #58 has a list of what this PR do. Thanks!

cuqi/testproblem/_testproblem.py
```diff
@@ -419,25 +419,40 @@ class Poisson_1D(BayesianProblem):
...
+
+    imap : lambda function
+        | Inverse of KL map.
 
     SNR : int
-        Signal-to-noise ratio
+        | Signal-to-noise ratio
+
+    n_steps: int, default 3
+        | Number of steps for the "Step" field.
```
nabriis, line 452: I think this parameter should be set via field_params instead. This is how the other geometries/fields get their parameters set.

cuqi/testproblem/_testproblem.py
```diff
@@ -419,25 +419,40 @@ class Poisson_1D(BayesianProblem):
...
+
+    observation_nodes : lambda function
```
nabriis, line 454: Would observation_map be more clear?

nabriis (changes requested):
Hi @amal-ghamdi,

Thanks for adding this new updated functionality to the two testproblems. Everything looks good, I just have one comment with respect to how the number of steps in the step expansion should be defined.

I think it should be as part of the field_params keyword

cuqi/testproblem/_testproblem.py
amal-ghamdi: Thank you @nabriis for your timely review! 

I think your suggestion is a very good idea but I see a complication here that might require further updates in `KLExpantion`, `KLExpansionFull` and `CustomKL`

For example: `KLExpansion_Full` initializer (and similarly the other 2 KL geometries) expect that params is a dictionary. 

```python
    def __init__(self, grid, params, axis_labels=['x'],**kwargs):

        super().__init__(grid, axis_labels,**kwargs)

        cl = params['cor_len']
        tau2 = 1./cl/cl
        nu = params["nu"]
```

while for example for `StepExpansion` Geometry, the initializer signature is:

```python
def __init__(self, grid, n_steps=3, fun2par_projection='mean', **kwargs):
```

my initial thought is that params in KL geometries should not be a dictionary and should be args or kwargs, for example 

```python
    def __init__(self, grid, nu, cor_len, axis_labels=['x'],**kwargs):
```

Then in Heat_1D one can pass a dictionary param_fields which will be passed with double asterisks `**param_fields` to the underlying geometry (in this case it will work whether the underlying geometry is StepExpansion or KL). 

We did something similar in, for example, plot_envelope parameters in method Samples.plot_ci 

We can talk about this in the stand-up

cuqi/testproblem/_testproblem.py
amal-ghamdi: I am not sure what is the best name here. I wanted to emphasize that this map takes the grid as input and spits out (subdomain) of the grid, do you think something like `observation_grid_map` is good, or we can use `observation_map`. My concern about the latter is that it might be confused with a general map (similar to Heat1D parameters: map and imap), what do you think? suggest?

jakobsj:
I believe the two of you are progressing this - please let me know if/when you'd like me to take a look.

cuqi/testproblem/_testproblem.py
nabriis: We discussed in standup to switch the KL stuff to keywords. Hopefully wont take too long

cuqi/testproblem/_testproblem.py
nabriis: I like observation grid map. Makes it a bit more clear

cuqi/testproblem/_testproblem.py
amal-ghamdi: Thank you @nabriis  for your feedback! I implemented the change we discussed today. To document the KL geometries parameters, I needed to write the expansion formula which was not straight forward to find. Now the user can look at where the parameters are used in the formula, for example, `cor_len`. 

To write the formulas I referred to a wiki article that discuss the idst (inverse discrete sign transform) https://en.wikipedia.org/wiki/Discrete_sine_transform. But most importantly, I wrote 2 methods in test_geometry test module that implements exactly what KLExpantion and KLExpansion_Full does (the par2fun method) but without using the spicy method `idst` that are used in  `KLExpantion` and `KLExpansion_Full`. The purpose of writing these two methods (that I am also using in tests): `_inverse_sin_discrete_transform_KLFull` and `_inverse_sin_discrete_transform_KL` is that they corresponds to the math described in the KL geometries documentation (so to insure the math is written correctly)

cuqi/testproblem/_testproblem.py
amal-ghamdi: Great, I use this name now

cuqi/testproblem/_testproblem.py
nabriis: Wow! This seems really good. I will have a closer look tomorrow.

amal-ghamdi: Great, thanks!

tests/test_geometry.py
```diff
@@ -174,3 +174,92 @@ def test_stepExpansion_fun2par(projection, func):
...
+    decay_rate = 2.5
+    normalizer = 12
+    geom = cuqi.geometry.KLExpansion(grid,
+                                     decay_rate=decay_rate,
+				     normalizer=normalizer)
+    p = np.random.randn(N)
+    f_geom = geom.par2fun(p)
+    f_expected = _inverse_sin_discrete_transform_KL(p, N, decay_rate, normalizer)
+
+    assert np.allclose(f_geom, f_expected)
```
nabriis, line 191: Nice!

nabriis (commented):
Thanks @amal-ghamdi, The PR looks really good. I had some minor suggestions to docs. But main code is great.

cuqi/geometry/_geometry.py
```diff
@@ -581,22 +581,52 @@ def __eq__(self, obj):
...
-    alpha = sum_i p * (1/i)^decay * sin(i*L*x/pi)
-    L is the length of the interval
-    '''
+    """
+    Class representation of the random field in the sine basis
+
+    .. math::
+        f = \sum_{i=0}^{N-2} (\\frac{1}{(i+1)^\\text{decay}\\text{normalizer}})  p_i \\, sin(\\frac{\\pi}{N}(i+1)(K+\\frac{1}{2})) 
+        
+        + \\frac{(-1)^K}{2}(\\frac{1}{(N)^\\text{decay}\\text{normalizer}})  p_{N-1}
```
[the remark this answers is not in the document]
amal-ghamdi: Thank you @nabriis for your review and very helpful comments! The grid is not used in the `KLExpansion` code and I think this is because the discrete sine transform is used. I think this is still makes sense, the only thing here is I think the grid needs to be regular for this idea to make sense, that's why I added `regular` in the grid description (I can also possibly add a check and error if the grid is not).

cuqi/geometry/_geometry.py
```diff
@@ -610,25 +640,55 @@ def fun2par(self,funvals):
...
-    class representation of the random field in  the sine basis
-    alpha = sum_i p * (1/i)^decay * sin(i*L*x/pi)
-    L is the length of the interval
+    Class representation of the random field in the sine basis
+
+    .. math::
+        \\tau = \\frac{1}{\\text{cor_len}^2}
+
+    .. math::
+        \\gamma = \\text{nu}+1
```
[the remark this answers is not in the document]
amal-ghamdi: Great, I did that. 

Yes they are very similar, I think the main difference is that the `i` is squared in the full KL. Also, the par2fun in the KL_expansion_Full can accept m (0<m<N) parameters. I can see that we can merge these in the future into one. 

I see this PR as an intermediate step, by documenting what is implemented and giving the user more control but surely the KL geometries could benefit from a future refactor.

amal-ghamdi (commented):
Thank you @nabriis for your very helpful review and edits. I worked on the suggestions you mentioned and I appreciate if you can take a second look (or a third :)).

cuqi/geometry/_geometry.py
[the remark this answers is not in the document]
nabriis: Ok. Sounds good to me.

nabriis (approved):
LGTM! Thanks for adding this.

amal-ghamdi:
Hi @jakobsj, Nicolai just approved this PR. I appreciate it if you can take a look. Thank you.

cuqi/geometry/_geometry.py
```diff
@@ -610,25 +648,58 @@ def fun2par(self,funvals):
...
+        The correlation length of the random field.
+
+    nu : float, default 2.5
+        Smoothness parameter of the random field.
+
+    std : float, default 1.0
+        Standard deviation of the random field.
     '''
     
     # init function defining paramters for the KL expansion
```
jakobsj, line 687:
```suggestion
    # init function defining parameters for the KL expansion
```

cuqi/geometry/_geometry.py
```diff
@@ -610,25 +648,58 @@ def fun2par(self,funvals):
...
-    def __init__(self, grid, params, axis_labels=['x'],**kwargs):
+    def __init__(self, grid, std=1.0, cor_len=0.2, nu=3.0, axis_labels=['x'],**kwargs):
```
jakobsj, line 688:
```suggestion
    def __init__(self, grid, std=1.0, cor_len=0.2, nu=3.0, axis_labels=['x'], **kwargs):
```

cuqi/geometry/_geometry.py
```diff
@@ -610,25 +648,58 @@ def fun2par(self,funvals):
...
 
         super().__init__(grid, axis_labels,**kwargs)
```
jakobsj, line 690:
```suggestion
        super().__init__(grid, axis_labels, **kwargs)
```

jakobsj: Several more like this, please search and replace.

cuqi/geometry/_geometry.py
```diff
@@ -581,22 +581,60 @@ def __eq__(self, obj):
...
+        One dimensional regular grid on which the random field is defined.
+
+    decay_rate : float, default 2.5
+        The decay rate of the basis functions.
+
+    normalizer : float, default 1.0
+        A factor of the basis functions shown in the formula above.
+    """
     
     # init function defining paramters for the KL expansion
```
jakobsj, line 617:
```suggestion
    # init function defining parameters for the KL expansion
```

jakobsj (changes requested):
Thank you @amal-ghamdi , lots of nice fixes here.  I see @nabriis did a very thorough review and I've taken a more superficial look through and don't see any issues (only added a couple of tiny typos etc) so no need to rereview.

amal-ghamdi:
Thank you @jakobsj for your review and catching the typos! Just a technicality here, I believe the merge is blocked pending your approval.

jakobsj (approved):

PR #63, opened 2022-09-07
Add default geometry on init of CUQIarray

### LINKED ISSUE #51: CUQIarray does not provide a default geometry

jakobsj:
Fixes #51


nabriis (commented):
Hi @jakobsj, thanks for adding this! It looks good. I just have one question.

What happens when ndarray has ndim>1? Maybe this relates to the 2D+ default geometry we have been talking about. Perhaps we can add a test to catch the behavior we want with that so we can keep track

nabriis (approved):

amal-ghamdi (commented):
Thank you @jakobsj for adding this feature! I just added the issue we talked about today #72.

PR #65, opened 2022-09-07
Sprint 11 Remove allow reduce flag for joint distribution

### LINKED ISSUE #64: Permanently allow joint distribution to reduce to a single density
Currently we have a hidden flag `_allow_reduce` that can be set to allow the joint distribution to reduce to a single density if possible. We have agreed to simply have this always happen.

nabriis:
Closes #64 

The edit in tests is because the Joint reduced to Posterior, which was not something that was aimed to test.

### DIFF
--- a/cuqi/distribution/_joint_distribution.py
+++ b/cuqi/distribution/_joint_distribution.py
@@ -67,7 +67,6 @@ class JointDistribution:
             raise ValueError("All densities must have unique names.")
 
         self._densities = list(densities)
-        self._allow_reduce = False # Hack to allow conditioning to reduce a joint distribution to a single density
 
         # Make sure every parameter has a distribution (prior)
         cond_vars = self._get_conditioning_variables()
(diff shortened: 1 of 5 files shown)


nabriis:
Hi @jakobsj, The build failed because the dev tutorial was trying to do something with the `_allow_reduce` flag that I missed. Its now fixed.

amal-ghamdi (approved):
I think this is a nice update that simplifies things (at least until samplers are able to work with JointDistribution). 

One comment, do you think keeping track of this reduction in the repr method makes sense?

For example when ever _reduce_to_single_density changes the object type, it adds a way to update the new object repr to show something like this

```
JointDistribution auto-converted to Posterior

Posterior(
    Equation:
        p(x|y) ∝ p(x)L(x|y)
    Densities:
        x ~ CUQI GaussianCov.
        y ~ CUQI GaussianCov Likelihood function. Parameters ['x'].
)
```

Just a thought.

amal-ghamdi:
Thank you @nabriis this LGTM!, I just added one comment but I approved nonetheless.

jakobsj (approved):
Seems good.  Amal raises a good point. I wonder if keeping track of the origin of objects more generally is useful - immediately fully specified distributions arising from conditioning conditional distributions comes to mind. I'd suggest making an issue on this idea in order to progress this PR now.

nabriis:
> Seems good. Amal raises a good point. I wonder if keeping track of the origin of objects more generally is useful - immediately fully specified distributions arising from conditioning conditional distributions comes to mind. I'd suggest making an issue on this idea in order to progress this PR now.

Great added #67

PR #71, opened 2022-09-09
Gaussian rework FU

furibec:
tasks done. 

We should also remove the JointGaussianSqrtPrec, now that we are organizing everything.


nabriis:
Great! Lets get it into gaussian_rework. I will make some further minor edits and send it off to amal and Jakob for review.

With respect to JointGaussianSqrtPrec, lets leave it in there for now. Gaussian does not support that functionality (which is to create a Gaussian from multiple stacked sqrtprec matricies).

PR #73, opened 2022-09-11
Update BayesianProblem to work with JointDistribution

### LINKED ISSUE #48: Update BayesianProblem to match behavior of JointDistribution
Currently our BayesianProblem requires a likelihood + prior combination. However, we are going away from explicitly converting to likelihood by the user and instead passing data by conditioning the joint distribution.

We need to update BayesianProblem to match this behavior. As a starting point we can throw an error when there are more than 2 distributions (1 data dist, 1 prior) since none of the automatic sampler selection is tuned for more than that.

See also comments in https://github.com/CUQI-DTU/CUQIpy-demos/issues/8

nabriis:
Closes #48 

- [x] Updated BayesianProblem to use JointDistribution as the underlying data structure
- [x] Updated docs for BayesianProblem
- [x] Added many new unit tests to capture new behavior of BayesianProblem (previously only had integration tests)
- [x] Updated many demo files + docs to use new style for BayesianProblem (It is still possible to use the likelihood, prior style, but prior parameter name must match the parameter name of the likelihood! It is no longer assumed to match)
- [x] Updated all testsproblems to ensure the parameter names match.
- [x] Added type hints to all methods of BayesianProblem
- [x] Consider adding a repr to BP with JointDistribution print + infostring?

Edit: This will likely cause some minor conflicts with #39. They should be easily handled.


nabriis:
Hi @jakobsj, @amal-ghamdi. I added both of you for review on this PR. Let me know at stand-up if there are any questions or if you need to go through it together.

demos/demo00_MinimalExample.py
```diff
@@ -4,49 +4,46 @@
...
-M  = LinearModel(forward=A)                                          #Model for inverse problem
-P  = Cauchy_diff(location=np.zeros(n),scale=0.05,bc_type='neumann')  #Prior distribution
-L  = Gaussian(mean=M,std=0.05).to_likelihood(data=b)                 #Likelihood function
-IP = BayesianProblem(likelihood=L, prior=P)                          #Bayesian model for inverse problem
-IP.UQ(exact=x_exact)                                                 #Perform UQ on inverse problem
+A  = LinearModel(Amat)                                  # y = Ax. Model for inverse problem
+x  = Laplace_diff(np.zeros(n), 0.01, bc_type='zero')    # x ~ Laplace_diff(0,0.01), Zero BC
+y  = Gaussian(A@x, 0.05)                                # y ~ N(Ax,0.05)
+IP = BayesianProblem(y, x).set_data(y=y_data)           # Bayesian problem given observed data
+IP.UQ(exact=x_exact)                                    # Run UQ analysis
```
amal-ghamdi, line 40: May be can return in samples for illustration 
```suggestion
samples = IP.UQ(exact=x_exact)                            # Run UQ analysis
```

demos/demo00_MinimalExample.py
```diff
@@ -4,49 +4,46 @@
...
-prior      = Cauchy_diff(np.zeros(n),0.05,'neumann')    #Prior distribution
-likelihood = Gaussian(model,0.05).to_likelihood(data=b) #Likelihood function
-IP         = BayesianProblem(likelihood,prior)          #Bayesian model for inverse problem
-IP.UQ()                                                 #Perform UQ on inverse problem        
\ No newline at end of file
+A  = LinearModel(Amat)                                  # y = Ax. Model for inverse problem
+x  = Cauchy_diff(np.zeros(n), 0.01, bc_type='zero')     # x ~ Cauchy_diff(0,0.01), Zero BC
+y  = Gaussian(A@x, 0.05)                                # y ~ N(Ax,0.05)
+IP = BayesianProblem(y, x).set_data(y=y_data)           # Bayesian problem given observed data
+IP.UQ(exact=x_exact)                                    # Run UQ analysis
```
amal-ghamdi, line 49: Same here, may be can return in samples for illustration 
```suggestion
samples = IP.UQ(exact=x_exact)                            # Run UQ analysis
```

tests/test_problem.py
```diff
@@ -0,0 +1,134 @@
...
+    # The tasks should be delegated to Posterior or JointDistribution
+    assert BP.posterior.prior.geometry == x.geometry # prior geometry is not changed
+    assert BP.posterior.likelihood.distribution.geometry == y.geometry # likelihood geometry is not changed
+    assert BP.data.geometry == y_obs.geometry # data geometry is not changed
+    assert BP.model == A # model geometry is not changed
+    
+    # Check that BayesianProblem correctly infers the geometry of the posterior (from model)
+    assert BP.posterior.geometry == domain_geometry
+    assert BP.sample_posterior(10).geometry == domain_geometry
+    #assert BP.sample_prior(10).geometry == domain_geometry # This fails because prior is sampled directly. TODO (should be domain_geometry)
```
amal-ghamdi, line 124: We mentioned in out discussion that we want to create an issue to capture geometries inconsistencies. Here is some documentation that can be converted to an issue.

For example:
- BP.sample_prior(10) generates samples with default geometry if prior has default geometry while the model has non-default geometry as domain geometry [sample prior should use domain geometry].
- Staked distribution: if two likelihood defined for the same prior, the samples generated from the stacked distribution will not be assigned the prior geometry and will be assigned default geometry instead [need to be assigned the prior geometry].

amal-ghamdi (changes requested):
Thank you @nabriis for the PR and adding many nice tests and updates! Now the BP is catching up with JointDistribution 👍 . I have few minor comments (mainly documentation) and one comment about the method set_data

nabriis:
Thanks @amal-ghamdi, since these comments are fairly minor, I will wait for @jakobsj comments before commiting any changes

README.md
```diff
@@ -24,24 +24,25 @@ from cuqi.data import grains
...
 
-# Step 3: Likelihood
-likelihood = GaussianCov(mean=model, cov=0.0036**2).to_likelihood(data)
+# Step 3: Likelihood, y ~ N(Ax, 0.0036^2)
+y = GaussianCov(mean=A@x, cov=0.0036**2)
 
-# Step 4: Posterior samples
-samples = BayesianProblem(likelihood, prior).sample_posterior(200)
+# Step 4: Set up Bayesian problem and sample posterior
+BP = BayesianProblem(y, x).set_data(y=y_data)
```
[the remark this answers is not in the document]
jakobsj: Great comment Amal!  I didn't think about this before, but completely agree that `set_data` having to return the BP (self) is a bit confusing. The syntax
```
BP.y = y_data
``` 
would be very consistent with the single-line (*densities, **data) syntax - is this possible? If so, `set_data` is probably redundant.  

The main strength of having `set_data` I think is that it makes it clearer that we are specifying observed data for a variable, whereas `BP.y = y1` could potentially be misunderstood to be replacing the random variable `y` given to the Bayesian Problem by `y1` (whatever that might be).

cuqi/problem/_problem.py
```diff
@@ -1,64 +1,65 @@
...
-    likelihood : Likelihood
-        The likelihood function.
+    \*densities: Density
+        The densities that represent the Bayesian Problem
+        Each density is passed as comma-separated arguments.
+        Can be Distribution, Likelihood etc.
 
-    prior : Distribution
-        The prior distribution.
+    \**data: ndarray, Optional
```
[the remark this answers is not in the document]
jakobsj: Agree.

cuqi/problem/_problem.py
```diff
@@ -1,64 +1,65 @@
...
+        Any potential observed data.
+        Data should be passed as keyword arguments.
+        Data can also be set later using the :meth:`set_data` method.
```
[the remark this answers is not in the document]
jakobsj: Agree (pending decision on whether to have `set_data` method or use syntax `BP.y = y_data`)

jakobsj (commented):
I agree with the comments by @amal-ghamdi - the main question is what is raised in 
https://github.com/CUQI-DTU/CUQIpy/pull/73#discussion_r968357072
ie whether we want the `set_data` method after all.
"There should be one– and preferably only one –obvious way to do it." (Zen of Python)

README.md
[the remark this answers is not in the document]
nabriis: After our discussion today we decided to stick with the `set_data` method that returns self (so it can be used either in the same line or in the line below). We agreed that BP is different than JointDistribution and so it need not follow exactly the same notation. BP is the non-expert interface, so in fact it should likely behave in a more user friendly way.

tests/test_problem.py
nabriis: #79

amal-ghamdi (approved):
Thank you @nabriis for the updates. LGTM!

cuqi/problem/_problem.py
```diff
@@ -76,54 +135,60 @@ def get_components(cls, **kwargs):
...
+
+    def set_data(self, **kwargs) -> BayesianProblem:
+        """ Set the data of the problem. This conditions the underlying joint distribution on the data. """
+        if not isinstance(self._target, JointDistribution):
+            raise ValueError("Unable to set data for this problem. Maybe data is already set?")
+        self._target = self._target(**kwargs)
+        return self
 
     @property
     def data(self):
```
jakobsj, line 149: Does this still work (is `likelihood` well-defined)?

cuqi/problem/_problem.py
```diff
@@ -76,54 +135,60 @@ def get_components(cls, **kwargs):
...
         return self.likelihood.data
 
     @property
-    def likelihood(self):
+    def likelihood(self) -> Likelihood:
         """The likelihood function."""
-        return self._likelihood
-    
+        if not isinstance(self._target, Posterior):
+            raise ValueError(f"Unable to extract likelihood from this problem. Current target is: \n {self._target}")
```
jakobsj, line 157: Is the term "target" defined? Maybe mention it in the doc string.

cuqi/problem/_problem.py
```diff
@@ -300,6 +365,8 @@ def UQ(self, exact=None):
         else:
             samples.plot_ci(95)
 
+        return samples
+
     def _sampleLinearRTO(self,Ns, callback=None):
```
jakobsj, line 370:
```suggestion
    def _sampleLinearRTO(self, Ns, callback=None):
```

cuqi/testproblem/_testproblem.py
```diff
@@ -522,22 +532,23 @@ def __init__(self, dim=128, dim_obs=None, endpoint=1, source=lambda xs: 10*np.ex
...
         # Generate exact data
-        b_exact = model.forward(x_exact,is_par=False)
+        y_exact = model.forward(x_exact,is_par=False)
 
         # Add noise to data
-        sigma = np.linalg.norm(b_exact)/SNR
+        sigma = np.linalg.norm(y_exact)/SNR
         sigma2 = sigma*sigma # variance of the observation Gaussian noise
-        data = b_exact + np.random.normal( 0, sigma, b_exact.shape )
+        data = y_exact + np.random.normal( 0, sigma, y_exact.shape )
```
jakobsj, line 540:
```suggestion
        data = y_exact + np.random.normal(0, sigma, y_exact.shape)
```

cuqi/testproblem/_testproblem.py
```diff
@@ -657,21 +668,22 @@ def PDE_form(IC, t): return (Dxx, np.zeros(N), IC)
                 x_exact = CUQIarray(x_exact, is_par=False, geometry=domain_geometry)
         #x_exact = 100*grid_domain*np.exp(-5*grid_domain)*np.sin(endpoint-grid_domain)
         # Generate exact data
-        b_exact = model.forward(x_exact,is_par=False)
+        y_exact = model.forward(x_exact,is_par=False)
```
jakobsj, line 671:
```suggestion
        y_exact = model.forward(x_exact, is_par=False)
```

cuqi/testproblem/_testproblem.py
```diff
@@ -657,21 +668,22 @@ def PDE_form(IC, t): return (Dxx, np.zeros(N), IC)
...
         # Add noise to data
-        sigma = np.linalg.norm(b_exact)/SNR
+        sigma = np.linalg.norm(y_exact)/SNR
         sigma2 = sigma*sigma # variance of the observation Gaussian noise
-        data = b_exact + np.random.normal( 0, sigma, b_exact.shape )
-
-        likelihood = cuqi.distribution.GaussianCov(model, sigma2*np.eye(model.range_dim)).to_likelihood(data)
-        prior = cuqi.distribution.GaussianCov(np.zeros(model.domain_dim), 1)
+        data = y_exact + np.random.normal( 0, sigma, y_exact.shape )
```
jakobsj, line 675:
```suggestion
        data = y_exact + np.random.normal(0, sigma, y_exact.shape)
```

cuqi/testproblem/_testproblem.py
```diff
@@ -881,27 +894,27 @@ def __init__(self, dim=128, endpoint=1, kernel=None, blur_size=48, field_type=No
...
         # Generate exact data
-        b_exact = model.forward(x_exact)
+        y_exact = model.forward(x_exact)
 
         # Add noise to data
-        sigma = np.linalg.norm(b_exact)/SNR
+        sigma = np.linalg.norm(y_exact)/SNR
         sigma2 = sigma*sigma # variance of the observation Gaussian noise
-        data = b_exact + np.random.normal( 0, sigma, b_exact.shape )
+        data = y_exact + np.random.normal( 0, sigma, y_exact.shape )
```
jakobsj, line 908:
```suggestion
        data = y_exact + np.random.normal(0, sigma, y_exact.shape)
```

cuqi/testproblem/_testproblem.py
```diff
@@ -1077,27 +1090,31 @@ def __init__(self,
         x_exact = CUQIarray(x_exact, is_par=True, geometry=domain_geometry)
 
         # Generate exact data (blurred)
-        b_exact = model @ x_exact
+        y_exact = model @ x_exact
```
jakobsj, line 1093:
```suggestion
        y_exact = model@x_exact
```

cuqi/problem/_problem.py
```diff
@@ -1,64 +1,123 @@
...
+        Can be Distribution, Likelihood etc.
+
+    \**data: ndarray, Optional
+        Any potential observed data. The data should be passed
+        as keyword arguments, where the keyword is the name of the
+        density that the data is associated with.
+        Data can alternatively be set using the :meth:`set_data` method,
+        after the problem has been initialized.
+
+    Example
```
jakobsj, line 45: This is a nice but comprehensive example. I think a super short - and incomplete - example would be helpful just to get the syntax across, e.g. something like
```
BayesianProblem(x, y, z, y=y_data)
```
or equivalently 
```
BayesianProblem(x, y, z).set_data(y=y_data)
```
for distributions x, y, z and observed data y_data for y.

jakobsj (changes requested):
Thank you @nabriis , I didn't find much, only a bit of spacing and one or two suggestions, otherwise looks good to me!

cuqi/problem/_problem.py
nabriis: Good point!

cuqi/problem/_problem.py
nabriis: Good suggestion. I added a basic syntax example.

jakobsj (approved):
Thanks, looks good!

PR #77, opened 2022-09-12
Add at most 10 ticks for Discrete plot

### LINKED ISSUE #76: Plot fewer ticks for Discrete geometry if size is large
In recent demos we try to plot discrete geometry with 128 objects. This causes some issues since the plotting adds too many ticks:

image

nabriis:
Closes #76


amal-ghamdi (approved):
LGTM!

jakobsj (approved):
And to me!

PR #82, opened 2022-09-13
Set default axis label to None

### LINKED ISSUE #50: Remove "x" from x-label when plotting by default
For example a plot like below does not make sense but is what happens by default if you plot data and exact solution.

Image

nabriis:
Closes #50


amal-ghamdi (commented):
LGTM! I think this update makes sense.

jakobsj (approved):
Looks good.

amal-ghamdi (commented):
LGTM!

amal-ghamdi (approved):

PR #83, opened 2022-09-13
Avoid sparse efficiency warning by doing the conversion ourselves

### LINKED ISSUE #81: Fix warning from GaussianCov sampling
It has to do with sparse efficiency. Should be an easy fix.

nabriis:
Without the warning :)

Closes #81


nabriis:
> LGTM! I approved. Just had one question, what if we convert the `self.sqrtprec` so we do not need to repeat the conversion process.

This would likely be more efficient, but since #39 is coming I went for the smallest possible change

jakobsj (approved):

PR #84, opened 2022-09-13
Sprint11 dist store var name geometry

### LINKED ISSUE #36: Allow distribution to store name as default variable name in Geometry
Plot labels for both trace plots show `v0`. Is it possible to get `d` and `l`? Perhaps also more descriptive titles on each of the two subplots, e.g., "Histogram: d"  and "Chain: d"?
image

_Originally posted by @jakobsj in https://github.com/CUQI-DTU/CUQIpy/pull/31#discussion_r957555068_

In dev-alpha the relevant code that added this feature was:
``` Python
            if hasattr(self, "_name"):
                name = str(self._name)
            else:
                name = 'v'
            value = [name+str(var) for var in range(value)]
```
in variable setter of Geometry

and
``` Python
            if self.name is not None:
                self._geometry._name = self.name
```
in geometry setter of Distribution.

Pretty ugly, we should find a better more robust way of doing it.

nabriis:
Closes #36


nabriis:
Hi @amal-ghamdi, I added a version of providing a variable name to Geometry class. In the future we may consider it a more public method, but for now it simply changes the default "v" to whatever the distribution name is (if it has one). 

Hence the variable for
``` Python
x = Gaussian(np.zeros(3), 1)
x.geometry.variables # Shows ["x0", "x1","x2"] instead of ["v0","v1","v2"]
```

Also if only one parameter, we do not index, i.e.
``` Python
x = Gaussian(0,1)
x.geometry.variables # Shows ["x"]
```

The user cannot modify this behavior

amal-ghamdi:
> Hi @amal-ghamdi, I added a version of providing a variable name to Geometry class. In the future we may consider it a more public method, but for now it simply changes the default "v" to whatever the distribution name is (if it has one).
> 
> Hence the variable for
> 
> ```python
> x = Gaussian(np.zeros(3), 1)
> print(x.name)
> x.geometry.variables # Shows ["x0", "x1","x2"] instead of ["v0","v1","v2"]
> ```
> 
> Also if only one parameter, we do not index, i.e.
> 
> ```python
> x = Gaussian(0,1)
> print(x.name)
> x.geometry.variables # Shows ["x"]
> ```
> 
> The user cannot modify this behavior

Thank you @nabriis, I think this works really nicely, and you can probably add your comment above as an issue to be addressed at a later time (giving the user control of the name)

amal-ghamdi (approved):
LGTM!

jakobsj (approved):

PR #86, opened 2022-09-14
Add visual only option to Image2D

nabriis:


cuqi/geometry/_geometry.py
```diff
@@ -437,11 +437,16 @@ class Image2D(Geometry):
         If order = 'C', the image is represented in row-major order.
         if order = 'F', the image is represented column-major order.
 
+    visual_only : bool, Default: False
```
amal-ghamdi, line 440: Do we need to add a check in CUQIarray when is_par is False 

```python
CUQIarray(np.ones(N,N), is_par=False, geom = geom)
```

where geom is an Image geometry for example.

cuqi/geometry/_geometry.py
```diff
@@ -452,16 +457,30 @@ def par_shape(self):
...
+        # Reshape to image (also for multiple parameter vectors). TODO: #327
+        image = vectors.reshape(self.fun_shape+(-1,), order=self.order) 
+        #Squeeze to return single image if only one parameter vector was given
+        image = image.squeeze()
+        return image
     
     def _plot(self, values, **kwargs):
+        # If only visual, we must convert value to image ourselves
+        if self.visual_only:
+            values = self._vector_to_image(values)
```
amal-ghamdi, line 482: does this work for multiple images?

cuqi/geometry/_geometry.py
```diff
@@ -452,16 +457,30 @@ def par_shape(self):
...
+        # Else, convert parameter vector into image
+        return self._vector_to_image(pars)
 
     def fun2par(self, funvals):
+        # If geometry is only used for visualization, do nothing
+        if self.visual_only: return funvals
+        # Else, convert image into parameter vector
         return funvals.ravel(order=self.order) #Maybe use reshape((self.dim,), order=self.order)
+
+    def _vector_to_image(self, vectors):
```
amal-ghamdi, line 471: nice helper function

amal-ghamdi (approved):
Thank you @nabriis. I think this is very nice option to add. I just have few comments, but I'm approving for now.

cuqi/geometry/_geometry.py
nabriis: Good point. I realized that we need to switch fun_shape depending on only_visual. Its now updated.

cuqi/geometry/_geometry.py
nabriis: Yes

amal-ghamdi (approved):
LGTM!

jakobsj (approved):

PR #108, opened 2022-09-22
Sprint12 pde related updates

### LINKED ISSUE #107: Add logic to cuqipy required to handle gradients in cuqipy-fenics

amal-ghamdi:
closes #107

### DIFF
--- a/cuqi/geometry/__init__.py
+++ b/cuqi/geometry/__init__.py
@@ -14,3 +14,13 @@ from ._geometry import (
     CustomKL,
     StepExpansion
 )
+
+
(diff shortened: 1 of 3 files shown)


nabriis (approved):
I think this is OK for now. It is kind of bad practise to put methods like that in the init, but I see why its needed.

I would suggest get rid of the `_idendity_geometries` ASAP after the workshop. Could you link the issue related to that? And perhaps add it to sprint 13 + set high priority?

amal-ghamdi:
Yes this makes sense. I am adding this issue that we already had in gitlab. #109

jakobsj (approved):

Issue #57, opened 2022-09-07
Add option to set parameter name during init of Model.


nabriis:
Currently the default parameter name for Model is "x". It is a bit cumbersome to change this to another name. One idea would be to initialize Model with a `parameter_name` attribute.

## Note
One point is that we can currently change the parameter name as follows
``` Python
y = cuqi.distribution.Gaussian(0, 1)
model_y = model(y) # Creates copy of model with parameter name y
```

nabriis:
With the latest PR the parameter name will be chosen according to the function parameter names. e.g.

``` Python
my_func = lambda z: z
Model(my_func, 1, 1) #gets the name y
```

PR #113, opened 2022-09-28
Sprint13 allow model other parameter names

### LINKED ISSUE #112: Allow models to actually work with other parameter names than x
Currently this kind of code fails

``` Python
import cuqi
y = cuqi.distribution.GaussianCov(0, 1)
model_s = cuqi.model.Model(lambda s: s, 1, 1)
model_y = model_s(y) # Switch name

# Only x works atm
assert model_s(s=1) == 1 
assert model_y(y=1) == 1
```

nabriis:
Closes #112, #57

### DIFF
--- a/cuqi/model/_model.py
+++ b/cuqi/model/_model.py
@@ -78,7 +78,7 @@ class Model(object):
             raise TypeError("The parameter 'domain_geometry' should be of type 'int' or 'cuqi.geometry.Geometry'.")
 
         # Store non_default_args of the forward operator for faster caching when checking for those arguments.
-        self._non_default_args = cuqi.utilities.get_non_default_args(self)
+        self._non_default_args = cuqi.utilities.get_non_default_args(self._forward_func)
 
     @property
     def domain_dim(self): 
@@ -184,9 +184,26 @@ class Model(object):
         x = self._input2fun(x, func_domain_geometry, is_par)
         out = func(x, **kwargs)
         return self._output2par(out, func_range_geometry, 
-                                    to_CUQIarray= (type(x) is CUQIarray)) 
+                                    to_CUQIarray= (type(x) is CUQIarray))
+
+    def _parse_args_add_to_kwargs(self, *args, **kwargs):
+        """ Private function that parses the input arguments of the model and adds them as keyword arguments matching the non default arguments of the forward function. """
+
+        if len(args) > 0:
+
+            if len(kwargs) > 0:
+                raise ValueError("The model input is specified both as positional and keyword arguments. This is not supported.")
+                
+            if len(args) != len(self._non_default_args):
+                raise ValueError("The number of positional arguments does not match the number of non-default arguments of the model.")
+            
+            # Add args to kwargs following the order of non_default_args
+            for idx, arg in enumerate(args):
+                kwargs[self._non_default_args[idx]] = arg
+
+        return kwargs
         
-    def forward(self, x, is_par=True ):
+    def forward(self, *args, is_par=True, **kwargs):
         """ Forward function of the model.
         
         Forward converts the input to function values (if needed) using the domain geometry of the model.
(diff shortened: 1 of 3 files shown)


amal-ghamdi (commented):
Thank you @nabriis for adding this great feature and the tests! I have only one comment. I was confused about the documentation of the utility function: 
```python
def get_non_default_args(func):
""" Returns the non-default arguments and kwargs from a callable function"""
```

While for example in the test here we have:
```python
model_z._non_default_args == ['z'] # the keyword argument not_used does not show up
```

does `get_non_default_args` actually returns the kwargs?

nabriis:
> Thank you @nabriis for adding this great feature and the tests! I have only one comment. I was confused about the documentation of the utility function:
> 
> ```python
> def get_non_default_args(func):
> """ Returns the non-default arguments and kwargs from a callable function"""
> ```
> 
> While for example in the test here we have:
> 
> ```python
> model_z._non_default_args == ['z'] # the keyword argument not_used does not show up
> ```
> 
> does `get_non_default_args` actually returns the kwargs?

Great question. It returns the arguments and keyword arguments that do not have a default value.

This is what was chosen a good while ago. Its important for partial functions, which are made by redefining a function with some partially specified arguments through the default mechanism. We use it when conditioning but only paetly specifying arguments of a function

amal-ghamdi:
> > Thank you @nabriis for adding this great feature and the tests! I have only one comment. I was confused about the documentation of the utility function:
> > ```python
> > def get_non_default_args(func):
> > """ Returns the non-default arguments and kwargs from a callable function"""
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > While for example in the test here we have:
> > ```python
> > model_z._non_default_args == ['z'] # the keyword argument not_used does not show up
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > does `get_non_default_args` actually returns the kwargs?
> 
> Great question. It returns the arguments and keyword arguments that do not have a default value.
> 
> This is what was chosen a good while ago. Its important for partial functions, which are made by redefining a function with some partially specified arguments through the default mechanism. We use it when conditioning but only partly specifying arguments of a function


Thank you @nabriis for the explanation. I am still not quite sure what  a keyword arguments that do not have a default value mean. Can you provide a simple example?

nabriis:
> > > Thank you @nabriis for adding this great feature and the tests! I have only one comment. I was confused about the documentation of the utility function:
> > > ```python
> > > def get_non_default_args(func):
> > > """ Returns the non-default arguments and kwargs from a callable function"""
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > While for example in the test here we have:
> > > ```python
> > > model_z._non_default_args == ['z'] # the keyword argument not_used does not show up
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > does `get_non_default_args` actually returns the kwargs?
> > 
> > 
> > Great question. It returns the arguments and keyword arguments that do not have a default value.
> > This is what was chosen a good while ago. Its important for partial functions, which are made by redefining a function with some partially specified arguments through the default mechanism. We use it when conditioning but only partly specifying arguments of a function
> 
> Thank you @nabriis for the explanation. I am still not quite sure what a keyword arguments that do not have a default value mean. Can you provide a simple example?

Yes, sorry for the confusion!

``` python 
def func(x, y, z=5):
    pass
```
Here x and y are non default arguments, and z is a default argument (because we do not have to set it)

amal-ghamdi (approved):
Thank you, it is just that I would think keyword arguments are always supplied with default values. But may I'm missing something. 

LGTM!

jakobsj (approved):
Nice fix. I read through the code but didn't run it. Appears to achieve what it claims to do. Great to see such comprehensive testing of the desired behavior :)

PR #119, opened 2022-10-10
preconditioned CGLS

furibec:
main function to be used with the horseshoe example later on

### DIFF
--- a/cuqi/solver/_solver.py
+++ b/cuqi/solver/_solver.py
@@ -274,8 +274,7 @@ class CGLS(object):
         # main loop
         k, flag, indefinite = 0, 0, 0
         while (k < self.maxit) and (flag == 0):
-            k += 1  
-
+            k += 1
             if self.explicitA:
                 q = self.A @ p
             else:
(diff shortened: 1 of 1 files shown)


nabriis:
> Hi @jakobsj, any chance we get this merge request approved? I just added the PCGLS functionality that I need to work on the horseshoe example.

Thanks for reaching out. I will look at it tonight Felipe

cuqi/solver/_solver.py
```diff
@@ -320,6 +326,137 @@ def solve(self):
...
+    See Bjorck (1996) - Numerical Methods for Least Squares Problems. Pag 294.
+
+    Parameters
+    ----------
+    A : ndarray or callable f(x,*args).
+    b : ndarray.
+    x0 : ndarray. Initial guess.
+    P : ndarray. Preconditioner array in sparse format.
+    maxit : The maximum number of iterations.
+    tol : The numerical tolerance for convergence checks.
```
nabriis, line 341: Could you add a few more comments here explaning what the parameters are. Specifically what is A, b and P. Show equation that is being solved.

Also please use the same format as in e.g. the [L_BGFS_B](https://github.com/CUQI-DTU/CUQIpy/blob/main/cuqi/solver/_solver.py#L17) class for parameters:
```
argument : type
    Description
```

cuqi/solver/_solver.py
```diff
@@ -320,6 +326,137 @@ def solve(self):
...
+        self.maxit = int(maxit)
+        self.tol = tol        
+        self.shift = shift
+        self.dim = len(x0)
+        if not callable(A):
+            self.explicitA = True
+        else:
+            self.explicitA = False
+        #
+        if self.dim < 1000:
```
nabriis, line 357: I would suggest using the same global config parameter for costly matrix computations: `cuqi.config.MAX_DIM_INV` instead of hard-coded 1000 (see here https://github.com/CUQI-DTU/CUQIpy/blob/main/cuqi/config.py#L4)

cuqi/solver/_solver.py
```diff
@@ -320,6 +326,137 @@ def solve(self):
...
+
+    def solve(self):
+        # initial state
+        x = self.x0.copy()
+        if self.explicitA:
+            r = self.b - (self.A @ x)
+            if self.explicitPinv:
+                s = self.Pinv.T @ (self.A.T @ r)
+            else:
+                s = self._apply_Pinv((self.A.T @ r), 2)
```
nabriis, line 376: I would suggest adding this kind of logic inside the function `_apply_Pinv`. It would simplify the main algorithm and make it easier to replace/change the functionality in the future.

cuqi/solver/_solver.py
```diff
@@ -320,6 +326,137 @@ def solve(self):
...
+            k += 1
+            #
+            if self.explicitPinv:
+                t = self.Pinv @ p
+            else:
+                t = self._apply_Pinv(p, 1)
+            if self.explicitA:
+                q = self.A @ t
+            else:
+                q = self.A(t, 1)
```
nabriis, line 402: Similar to how you created a function for `_apply_Pinv` I would suggest adding one for `_apply_A`.

cuqi/solver/_solver.py
```diff
@@ -320,6 +326,137 @@ def solve(self):
...
+            flag = 3   # Matrix (A'*A + delta*L) seems to be singular or indefinite
+            ValueError('\n Negative curvature detected !')  
+        if shrink <= np.sqrt(self.tol):
+            flag = 4   # Instability likely: (A'*A + delta*L) indefinite and NORM(X) decreased
+            ValueError('\n Instability likely !') 
+
+        return x, k
+
+    def _apply_Pinv(self, x, flag):
+        if has_cholmod:
```
nabriis, line 447: For future reference could you please add a brief description in the docstring of this method what it does?

nabriis (changes requested):
Hi @furibec, Thank you for adding this nice new PCGLS method! I had some refactoring comments I hope you agree with. Let me know if you have any questions.

furibec:
Hi @nabriis , thanks a lot for checking this function. I incorporated your comments, I liked the idea of creating the apply_A, and putting the "is" logic inside the method functions. Let me know if you have further comments.

nabriis (approved):
Thanks @furibec for the quick response! Looks good to me! Ill add 2nd reviewer.

cuqi/solver/_solver.py
```diff
@@ -320,6 +327,136 @@ def solve(self):
...
+        if self.dim < config.MAX_DIM_INV:
+            self.explicitPinv = True
+            Pinv = spa.linalg.inv(P)
+        else:
+            self.explicitPinv = False
+            Pinv = None # we do cholesky.solve or sparse.solve with P
+            if has_cholmod:
+                # turn P into a cholesky object
+                P = cholesky(P, ordering_method='natural')
+        self.Pinv = Pinv # inverse of the preconditioner as a ndarray or function
```
amal-ghamdi, line 373: Normally, if there is dependency in the attributes (for example Pinv depends on P) we create setters and getters for them so that the depndenant property can be inferred correctly if the original property changed. Or we can also  make those properties private to indicate that the user is not supposed to change them after initialization. For the scenario when the user create the solver: `mysolver = PCGLS(...)` then afterwards they do something like: `my solver.A = ...`,  `self.explicitPinv` needs to be updated which can be done through a setter. However, I suggest for now we make all such properties private (e.g.: `self._A`, `self._x0`, `self._P`,`self._Pinv`, `self._explicitPinv`,  `self._explicitA`, `self._dim`). And we can provide getters if needed. 

for example for A, in the initializer:
 self._A = A

and if needed we add a property
```python
@property
def A(self):
    return self._A
```
Here the user cannot set A but can retrieve A after initialization.

amal-ghamdi (commented):
Thank you @furibec for this great contribution! I just have one comment about class attribute names.

cuqi/solver/_solver.py
furibec: Hi @amal-ghamdi, thanks for the comment. As you suggested, I made the properties private for the moment since it is the easiest solution so I can focus on the horseshoe. I am not really familiar with coding properties, setters and getters, so I guess this has to be implemented for all classes in solver.py at some future re-work.

amal-ghamdi (approved):
Looks great. Thank you @furibec! I think as you suggested adding properties at solver level is a good idea.

PR #121, opened 2022-10-14
Sprint11 multiple likelihoods

### LINKED ISSUE #120: Add a demo showing a bayesian model with multiple likelihoods

amal-ghamdi:
closes #120


demos/howtos/multiple_likelihoods.py
```diff
@@ -0,0 +1,180 @@
...
+# -----------------------
+
+observation_grid_map2 = None
+if set_up == "multi_observation":
+        # Observe on the second half of the domain
+	observation_grid_map2 = lambda x: x[np.where(x>=.5)]
+
+# The source term signal
+if set_up == "multi_source":
+	source2 = lambda xs: 20*np.sin(2*xs)+20.1
```
jakobsj, line 82: It took me a little while to determine what the difference between the two likelihoods were in the multi source case, as the code reads almost the same, only a "2*" hidden inside the sine. The plots of the data for the two different sources look almost identical. Could we make the difference clearer somehow? For example could we make plots of the two source terms?

demos/howtos/multiple_likelihoods.py
```diff
@@ -0,0 +1,180 @@
...
+y1 = cuqi.distribution.Gaussian(mean=model1, std=sigma_noise1,
+                                corrmat=np.eye(model1.range_dim), geometry=model1.range_geometry)
+y2 = cuqi.distribution.Gaussian(mean=model2, std=sigma_noise2,
+                                corrmat=np.eye(model2.range_dim), geometry=model2.range_geometry)
+
+# %%
+# Create the posterior (multiple likelihoods)
+# -------------------------------------------
+
+z_joint = cuqi.distribution.JointDistribution(x,y1,y2)(y1=data1, y2=data2)._as_stacked() # _as_stacked() is needed to stack the random variables but it is a temporary hack that will be not needed in the future.
```
jakobsj, line 129: Is there an issue to remove the need for _as_stacked?

demos/howtos/multiple_likelihoods.py
```diff
@@ -0,0 +1,180 @@
...
+# %%
+# Sample from the posterior (multiple likelihoods)
+# ------------------------------------------------
+
+# Sample from the posterior
+sampler = cuqi.sampler.MetropolisHastings(z_joint)
+samples = sampler.sample_adapt(5000)
+
+# Set the samples geometry
+samples.geometry=x.geometry # this is will be not needed in the future as samples geometry will be automatically determined.
```
jakobsj, line 140: Is there an issue for this?

jakobsj (changes requested):
Thank you very @amal-ghamdi for this very nice and clear demonstration of how to use multiple likelihoods in CUQIpy. I like both examples a lot, they are easy to follow and nicely convey the concept. A couple of points came up as I worked through:

1. I continue to find the line between tutorial and how-to rather blurred. There are various definitions around, for example https://diataxis.fr/tutorials-how-to/ .   The present case I would say could probably fit in either category, since it explains how to achieve the desired task but also gives a complete workflow and makes several comparisons, on the other does not give so much explanation of what is observed (to be clear, I like the comparisons as they give context, but for a howto maybe context is not to be included?) So I see various paths - one could merge to a single category (e.g. only have, say, tutorials, where some can be more detailed than others), or one could try to distinguish more strictly for example make definitions/recipes for what goes in a tutorial and in a howto. I assume a more strict definition of howtos would tend to produce very short, concise examples with little to no context/comparisons.

2. Currently the two-likelihood bit is done first, followed by the two single likehood one after the other. Might it make sense to move it to being last as it is the most advanced and key concept of this demo? Then the more basic common case of using single likelihoods act as stepping stones toward the "grand finale" of how to use multiple at the same time?  

3. Related to both of the above points, at the end of the notebook I was a bit unsure why it ended there, what the main takeaway was, what to be seen from plot_ci plots and ess values? One could add a bit more explanation (ie making it more tutorial like...) of what is seen in the plots ie how using multiple likelihoods improves something - in fact in the multi source case the multi likelihood sampling did not immediately appear better to me.  I think reorganizing as suggested in point 2 would help make it clearer why the demo ends there, alternatively explain in comments early on that the multi likelihood case will first be demonstrated and then compared with single likelihood cases.

amal-ghamdi:
Thank you @jakobsj for your thorough review and very helpful comments. I addressed all the points you brought up, mainly:

1- I moved this script into the tutorials section adding some more explanation of what we do and achieve 
2- Reorder the numerical experiments to make the single likelihood cases comes first
3- add some discussion of the obtained results.

I also added two issues #123 and #124 for eliminating the need for using `as_stacked` and setting the geometry manually. 

In regards to the source term, I realized I did not pass it to the test problem :), so thank you for brining this up. Now the data will look different in both cases so there is probably no need for plotting the source term.

demos/howtos/multiple_likelihoods.py
amal-ghamdi: This was a bug because I did not pass the source term to the testproblem. Now it is fixed :)

demos/howtos/multiple_likelihoods.py
amal-ghamdi: Issue #123 created

demos/howtos/multiple_likelihoods.py
amal-ghamdi: issue #124 created

jakobsj (changes requested):
Thank you @amal-ghamdi , your new version nicely addresses all my previous comments.  I like the flow and explanations and I think their amount is just right.  Two new requests:

1. Some of the lines are rather long. This makes them a bit hard to read in the editor, requiring some scrolling. The general recommendation (PEP8) is max 79 characters, I understand this is not a technical requirement but mainly for readability. And it's fine that some lines are slightly longer. So I would suggest to wrap the longer lines.

2. As you mentioned the display of the posterior is very informative, and it works nicely also in this case with multiple likelihood. I think it would be nice to print at least the joint multiple likelihood posterior (possibly also the single-likelihood ones)  and comment briefly on what is seen in the print, as that gives quite some insight into the similarities and differences of single and multi likelihood posteriors.

Otherwise I find it good to go!

jakobsj:
It is mainly lines containing comments that are long.

jakobsj (approved):
Thank you, looks great!

demos/tutorials/multiple_likelihoods.py
```diff
@@ -0,0 +1,268 @@
...
+model1, data1, problemInfo1 = cuqi.testproblem.Poisson_1D.get_components(dim=dim,
+                                                                         endpoint=endpoint,
+                                                                         field_type=field_type,
+                                                                         field_params={
+                                                                             'n_steps': n_steps},
+                                                                         observation_grid_map=observation_grid_map1,
+                                                                         exactSolution=x_exact,
+                                                                         source=source1,
+                                                                         SNR=SNR)
+
```
nabriis, line 85: This does not render nicely on the online docs as the line is too long. I would suggest using the black-style formatting.

We should set up a GitHub action to automtically apply black formatting, but until then you can use http://black.vercel.app
```suggestion
model1, data1, problemInfo1 = cuqi.testproblem.Poisson_1D.get_components(
    dim=dim,
    endpoint=endpoint,
    field_type=field_type,
    field_params={"n_steps": n_steps},
    observation_grid_map=observation_grid_map1,
    exactSolution=x_exact,
    source=source1,
    SNR=SNR,
)

```

demos/tutorials/multiple_likelihoods.py
```diff
@@ -0,0 +1,268 @@
...
+# %%
+# Create the prior
+# ----------------
+#
+# Create the prior for the Bayesian parameter `theta`, which is the expansion 
+# coefficients of the conductivity (or diffusivity) step function. We use a 
+# Gaussian prior.
+
+theta = cuqi.distribution.GaussianCov(np.zeros(
+	model1.domain_dim), 3*np.ones(model1.domain_dim), geometry=model1.domain_geometry)
```
nabriis, line 140:
```suggestion
theta = cuqi.distribution.GaussianCov(
    np.zeros(model1.domain_dim),
    3 * np.ones(model1.domain_dim),
    geometry=model1.domain_geometry,
)

```

nabriis (changes requested):
Hi @amal-ghamdi, I think the tutorial looks good. I had some minor comments on formatting.

amal-ghamdi:
> Hi @amal-ghamdi, I think the tutorial looks good. I had some minor comments on formatting.

Thank you @nabriis for the suggested updates! it looks much better now in the documentation page.

nabriis (changes requested):
Please rebase on main and use the new Gaussian :)

amal-ghamdi:
> Please rebase on main and use the new Gaussian :)

done 👍

nabriis (approved):
LGTM!

PR #122, opened 2022-10-18
Update funvals method of CUQIarray class

amal-ghamdi:
closes CUQI-DTU/CUQIpy-FEniCS#14

### DIFF
--- a/cuqi/samples/_samples.py
+++ b/cuqi/samples/_samples.py
@@ -73,9 +73,12 @@ class CUQIarray(np.ndarray):
             vals = self
 
         if isinstance(vals, np.ndarray):
-            return type(self)(vals,is_par=False,geometry=self.geometry) #vals.view(np.ndarray)
-        else: 
-            return vals  
+            if vals.dtype == np.dtype('O'):
+                return self.reshape(1)[0]
+            else:
+                return type(self)(vals,is_par=False,geometry=self.geometry) #vals.view(np.ndarray)
+        else:
+            return vals 
 
     @property
     def parameters(self):
--- a/cuqi/samples/_samples.py
+++ b/cuqi/samples/_samples.py
@@ -74,8 +74,13 @@ class CUQIarray(np.ndarray):
 
         if isinstance(vals, np.ndarray):
             if vals.dtype == np.dtype('O'):
+                # if vals is of type np.ndarray, but the data type of the array
+                # is object (e.g. FEniCS function), then extract the object and
+                # return it. reshape(1) is needed to convert the shape from
+                # () to (1,).
                 return self.reshape(1)[0]
             else:
+                # else, cast the np.ndarray to a CUQIarray
                 return type(self)(vals,is_par=False,geometry=self.geometry) #vals.view(np.ndarray)
         else:
             return vals 


amal-ghamdi:
Tests for this update are handled by CUQI-DTU/CUQIpy-FEniCS#14. These tests currently does not pass but once this PR is merged the tests in CUQI-FEniCS will be triggered again and supposedly will pass.

nabriis (commented):
Hi @amal-ghamdi, it seems to me that we are checking if an object was put into a numpy array and then extracting it again? Are we sure we want to support this behavior? It would be more clear for me if other parts of the code does not put objects into numpy arrays if they do not belong?

amal-ghamdi:
> Hi @amal-ghamdi, it seems to me that we are checking if an object was put into a numpy array and then extracting it again? Are we sure we want to support this behavior? It would be more clear for me if other parts of the code does not put objects into numpy arrays if they do not belong?
> 
> Edit: I see now that we want to support stuff like FEniCS objects that represent function values.
> 
> I think it is very mystical to put these into a numpy array (seems to be from par2fun?), but less mystical if its a CUQIarray. I wonder if we could only check for the CUQIarray case? (I know CUQIArray is currently subclassed from NumPy, but it may not be in the future)

Thank you @nabriis 

Yes I agree with you it is not optimal and I believe that subclassing CUQIarray, though very powerful, limits what type of objects can be held as a function value.

Now the current implementation of CUQIarray (unrelated to this PR) will wrap the passed value into numpy array automatically, e.g.

```python
u = dl.Function(V)
u_CUQIarray = cuqi.samples.CUQIarray(u, is_par=False, geometry=model.domain_geometry)
u_CUQIarray
```

will results in
```
CUQIarray: NumPy array wrapped with geometry.
---------------------------------------------

Geometry:
 FEniCSContinuous(101,)

Parameters:
 False

Array:
CUQIarray(Coefficient(FunctionSpace(Mesh(VectorElement(FiniteElement('Lagrange', interval, 1), dim=1), 0), FiniteElement('Lagrange', interval, 1)), 27),
          dtype=object)
```
And this is what cause two tests cases to fail:

```python
    # Test passing a CUQIarray containing dolfin function
     u = dl.Function(V)
     u_CUQIarray = cuqi.samples.CUQIarray(u, is_par=False, geometry=model.domain_geometry)
     y = model(u_CUQIarray)

     # Test passing a CUQIarray containing dolfin function wrapped in np.array
     # This is not recommended, but should work
     u = dl.Function(V)
     u_CUQIarray = cuqi.samples.CUQIarray(np.array(u, dtype='O'), is_par=False, geometry=model.domain_geometry)
     y = model(u_CUQIarray) 
```

May be I can look into the CUQIarray initializer and update it to not wrap objects in numpy array?

nabriis (approved):
Upon further thought I cannot figure a good way to test or capture this. I think it looks good to me :)

jakobsj (changes requested):
Good fix! The code is hard to understand without context, so please add comments to explain the purpose of each of the if statements (incl in which situation, perhaps an example, this code is required), as well as what is achieved by `np.dtype('O')`, `self.reshape(1)[0]` and `type(self)(vals,is_par=False,geometry=self.geometry)`.

amal-ghamdi:
Thank you @jakobsj for your helpful review! I updated the documentation in `funvals` I appreciate it if you can take a look at the updates.

jakobsj (approved):

PR #133, opened 2022-10-20
Publish pre-release directly to PyPI

### LINKED ISSUE #132: Add pre-releases to PyPI
This way the user can update to the latest version from main user the flag `--pre`. Similarly we can use pre-releases for our testing of plugins without relying on test pypi.

nabriis:
Closes #132


jakobsj (approved):
Seems fine!

PR #142, opened 2022-11-04
Sprint14 fd gradient

### LINKED ISSUE #143: Discussion: Enable FD approximation for posterior gradient (density in general)
Preliminary implementation in PR #142 

More to consider:

- [x]  Using FD approximation works with ULA and MALA, make sure it works for NUTS (especially when function space and parameter space dimensions are different) (issue added to address NUTS bug issue #157 )
- [x] Enable FD gradient for Density in general (can make parent method `gradient` in density that execute the FD logic if use_FD is enabled, then call subclass `_gradient` if use_FD is false )
- [x] Update the FD approximation method in utilities to accept CUQIarray
- [x] give user flexibility to specify FD method step size (also possibly user can pass the FD method itself. e.g. second order approximation) (Step can be controlled, but method not yet)
- [x] Decide if we want the accompanying demo to be in how-to or tutorial (It is a demo for now)

amal-ghamdi:
closes #143

### DIFF
--- a/cuqi/distribution/_distribution.py
+++ b/cuqi/distribution/_distribution.py
@@ -187,6 +187,12 @@ class Distribution(Density, ABC):
         """
         pass
 
+    def _gradient(self, *args, **kwargs):
+        raise NotImplementedError(
+            f"_gradient is not implemented for {self.__class__.__name__}. " +
+            "Consider setting the property use_FD to True to use finite " +
+            "differences to approximate the gradient.")
+
     def sample(self,N=1,*args,**kwargs):
         #Make sure all values are specified, if not give error
         #for key, value in vars(self).items():
--- a/cuqi/distribution/_distribution.py
+++ b/cuqi/distribution/_distribution.py
@@ -190,8 +190,9 @@ class Distribution(Density, ABC):
     def _gradient(self, *args, **kwargs):
         raise NotImplementedError(
             f"_gradient is not implemented for {self.__class__.__name__}. " +
-            "Consider setting the property use_FD to True to use finite " +
-            "differences to approximate the gradient.")
+            "Consider enabling finite difference approximation of the " +
+            f"gradient by calling the {self.__class__.__name__} method " +
+            "enable_FD().")
 
     def sample(self,N=1,*args,**kwargs):
         #Make sure all values are specified, if not give error
--- a/cuqi/likelihood/_likelihood.py
+++ b/cuqi/likelihood/_likelihood.py
@@ -182,8 +182,6 @@ class UserDefinedLikelihood(object):
         """Returns value of likelihood function"""
         return self.logpdf_func(*args, **kwargs)
 
-    #TODO: Change gradient to _gradient when this class is subclassed from
-    # Density.
     def gradient(self, *args, **kwargs):
         """Return gradient of likelihood function"""
         return self.gradient_func(*args, **kwargs)
--- a/cuqi/likelihood/_likelihood.py
+++ b/cuqi/likelihood/_likelihood.py
@@ -44,24 +44,14 @@ class Likelihood(Density):
         self.distribution.name = value
 
     @property
-    def use_FD(self):
-        """ Return use_FD of the likelihood from the underlying distribution """
-        return self.distribution.use_FD
-
-    @use_FD.setter
-    def use_FD(self, value):
-        """ Set use_FD of the underlying distribution of the likelihood """
-        self.distribution.use_FD = value
+    def FD_enabled(self):
+        """ Return FD_enabled of the likelihood from the underlying distribution """
+        return self.distribution.FD_enabled
 
     @property
     def FD_epsilon(self):
         """ Return FD_epsilon of the likelihood from the underlying distribution """
         return self.distribution.FD_epsilon
-
-    @FD_epsilon.setter
-    def FD_epsilon(self, value):
-        """ Set FD_epsilon of the underlying distribution of the likelihood """
-        self.distribution.FD_epsilon = value
 
     @property
     def _constant(self):
@@ -144,6 +134,15 @@ class Likelihood(Density):
     # Overload parent to add type hint.
     def __call__(self, *args, **kwargs) -> Union[Likelihood, EvaluatedDensity]:
         return super().__call__(*args, **kwargs)
+
+    def enable_FD(self, epsilon=1e-8):
+        """ Call enable_FD of the underlying distribution """
+        self.distribution.enable_FD(epsilon)
+
+    def disable_FD(self):
+        """ Call disable_FD of the underlying distribution """
+        self.distribution.disable_FD()
+
 
 class UserDefinedLikelihood(object):
     """ Class to wrap user-defined likelihood functions.
--- a/cuqi/likelihood/_likelihood.py
+++ b/cuqi/likelihood/_likelihood.py
@@ -54,14 +54,14 @@ class Likelihood(Density):
         self.distribution.use_FD = value
 
     @property
-    def FD_tol(self):
-        """ Return FD_tol of the likelihood from the underlying distribution """
-        return self.distribution.FD_tol
+    def FD_epsilon(self):
+        """ Return FD_epsilon of the likelihood from the underlying distribution """
+        return self.distribution.FD_epsilon
 
-    @FD_tol.setter
-    def FD_tol(self, value):
-        """ Set FD_tol of the underlying distribution of the likelihood """
-        self.distribution.FD_tol = value
+    @FD_epsilon.setter
+    def FD_epsilon(self, value):
+        """ Set FD_epsilon of the underlying distribution of the likelihood """
+        self.distribution.FD_epsilon = value
 
     @property
     def _constant(self):
--- a/cuqi/likelihood/_likelihood.py
+++ b/cuqi/likelihood/_likelihood.py
@@ -44,6 +44,26 @@ class Likelihood(Density):
         self.distribution.name = value
 
     @property
+    def use_FD(self):
+        """ Return use_FD of the likelihood from the underlying distribution """
+        return self.distribution.use_FD
+
+    @use_FD.setter
+    def use_FD(self, value):
+        """ Set use_FD of the underlying distribution of the likelihood """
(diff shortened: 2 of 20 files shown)


nabriis (changes requested):
We discussed the suggested changes in a meeting. Basic summary is to make the FD array type independant, and removing the init making it into a configure_FD method instead.

amal-ghamdi:
> We discussed the suggested changes in a meeting. Basic summary is to make the FD array type independant, and removing the init making it into a configure_FD method instead.

Thanks @nabriis for your feedback. I made the updates we talked about along with using FD method (utilities.approx_gradient) that is specific for scalar functions (since logd is a scalar valued function). It will be great if you can take a look.

cuqi/utilities/_utilities.py
```diff
@@ -173,18 +163,45 @@ def approx_derivative(func, wrt, direction=None, epsilon=np.sqrt(np.finfo(np.flo
...
+    step size `epsilon`."""
+
+    #TODO: use approx_derivative after making it data structure agnostic
+    #return approx_derivative(func, x, None, epsilon=epsilon)
+ 
+    if isinstance(x, Number):
+        return (func(x+epsilon) - func(x))/epsilon
+
+    FD_gradient = copy.deepcopy(x)*0.0
+    eps_vec = copy.deepcopy(x)*0.0
```
nabriis, line 199: Is the deepcopy needed? This can be rather slow. How about simply `x*0.0?`

cuqi/utilities/_utilities.py
```diff
@@ -173,18 +163,45 @@ def approx_derivative(func, wrt, direction=None, epsilon=np.sqrt(np.finfo(np.flo
...
+
+    for i, x_i in enumerate(x):
```
nabriis, line 201: Not sure if all arrays are enumerateble like that. Perhaps a simply len(x) and loop over that is better?

nabriis (approved):
Hi Amal. This looks good to me now. I have two minor suggestions you can look at. Otherwise it looks ready.

cuqi/utilities/_utilities.py
amal-ghamdi: I think this is a nice simplification 👍

cuqi/utilities/_utilities.py
amal-ghamdi: Good point, I ended up using infer_len because __len__ is also not necessarily implanted I think, based on array API specification: https://data-apis.org/array-api/latest/API_specification/index.html

nabriis: Great thanks!

demos/demo35_FD_gradient.py
```diff
@@ -0,0 +1,27 @@
...
+posterior = TP.posterior
+
+#%% Sample from the posterior using Metropolis-Hastings
+MH_sampler = cuqi.sampler.MetropolisHastings(posterior)
+MH_samples = MH_sampler.sample_adapt(1000)
+MH_samples.plot_ci(95,exact=TP.exactSolution)
+
+#%% Sample from the posterior using MALA
+plt.figure()
+posterior.enable_FD()
```
jakobsj, line 18: Would it be illustrative to make an attempt to call MALA without having enabled FD and seeing it fail? The script would not run, so another idea would be to put the attempt in a commented code line, and explain that had it been run it would have failed, because it requires gradients which we don't have off the bat, so need to enable the FD.

tests/test_likelihood.py
```diff
@@ -111,3 +111,30 @@ def test_likelihood_conditioning(dist, mean, cov, data):
...
+    likelihood = y(y=data)
+
+    # Compute exact gradient
+    g_exact = likelihood.gradient(x_i)
+
+    # Compute FD gradient
+    likelihood.enable_FD(1e-7)
+    g_FD = likelihood.gradient(x_i)
+
+    # Assert that the exact and FD gradient are close,
```
jakobsj, line 138: Nice test using that we can have both analytical and FD gradients!

cuqi/utilities/_utilities.py
```diff
@@ -173,18 +163,46 @@ def approx_derivative(func, wrt, direction=None, epsilon=np.sqrt(np.finfo(np.flo
     # corresponding geometry par2fun map is not identity
     # (e.g. funvalues=paramters**2), because wrt entries 
     # are interpreted as function value.
-    if isinstance(wrt, CUQIarray) or isinstance(direction, CUQIarray):
-        raise NotImplementedError("approx_derivative is not implemented"+
-                                  "for inputs of type CUQIarray")
+
```
jakobsj, line 166: Are the above comments still relevant? An error doesn't appear to be thrown anymore?

cuqi/utilities/_utilities.py
```diff
@@ -173,18 +163,46 @@ def approx_derivative(func, wrt, direction=None, epsilon=np.sqrt(np.finfo(np.flo
...
+    if isinstance(wrt, CUQIarray):
```
jakobsj, line 167: Please add more comments explaining the following ca 20 lines.

cuqi/utilities/_utilities.py
```diff
@@ -173,18 +163,46 @@ def approx_derivative(func, wrt, direction=None, epsilon=np.sqrt(np.finfo(np.flo
...
+        return Matr@direction
+
+def approx_gradient(func, x, epsilon= 0.000001):
+    """Approximates the gradient of callable scalar function `func` evaluated at point `x`. The approximation is done using finite differences with
+    step size `epsilon`."""
+
+    #TODO: use approx_derivative after making it data structure agnostic
+    #return approx_derivative(func, x, None, epsilon=epsilon)
+ 
+    if isinstance(x, Number):
```
jakobsj, line 195: Please add comments explaining the next ca 15 lines.

docs/conf.py
```diff
@@ -84,10 +84,17 @@
 
 
 html_logo = "../logo.png"
+
+#TODO: The option 'logo' is to be removed once pydata-sphinx-theme
```
jakobsj, line 88: Add issue (remove TODO?)

jakobsj (changes requested):
Thank you for this comprehensive PR adding finite different based gradient. I find it is in great shape and have added some minor comments and questions.  This issue it is aimed at, namely #143 mentions another issue to be addressed #157, is this covered by this PR or for another PR?

amal-ghamdi:
> 

Thank you @jakobsj  for the very helpful review and great suggestions! Issue #157 was created to handle this bug separately. It is not fixed as part of this PR.

cuqi/distribution/_distribution.py
```diff
@@ -187,6 +187,12 @@ def logpdf(self,x):
         """
         pass
 
+    def _gradient(self, *args, **kwargs):
+        raise NotImplementedError(
+            f"_gradient is not implemented for {self.__class__.__name__}. " +
+            "Consider setting the property use_FD to True to use finite " +
```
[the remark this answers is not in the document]
amal-ghamdi: good catch! I updated it.

cuqi/likelihood/_likelihood.py
```diff
@@ -163,6 +182,8 @@ def logd(self, *args, **kwargs):
         """Returns value of likelihood function"""
         return self.logpdf_func(*args, **kwargs)
 
+    #TODO: Change gradient to _gradient when this class is subclassed from
```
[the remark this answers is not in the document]
amal-ghamdi: Added issue #184 and removed the TODO

cuqi/samples/_samples.py
```diff
@@ -92,7 +92,15 @@ def parameters(self):
         else:
             vals = self
         return type(self)(vals,is_par=True,geometry=self.geometry)
-    
+
+    def to_numpy(self):
```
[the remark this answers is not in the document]
amal-ghamdi: This is a helper function that access the numpy array in CUQIarray. I found it a bit indirect how to return the numpy array itself from the CUQIarray so I added this method. It could be useful for the user. I do not think self.array works to retrieve the numpy array.

I added some documentation to clarify the effect of is_par.

cuqi/utilities/_utilities.py
```diff
@@ -173,18 +163,46 @@ def approx_derivative(func, wrt, direction=None, epsilon=np.sqrt(np.finfo(np.flo
...
     else:
-        return Matr@direction
\ No newline at end of file
+        return Matr@direction
+
+def approx_gradient(func, x, epsilon= 0.000001):
+    """Approximates the gradient of callable scalar function `func` evaluated at point `x`. The approximation is done using finite differences with
+    step size `epsilon`."""
+
+    #TODO: use approx_derivative after making it data structure agnostic
```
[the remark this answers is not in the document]
amal-ghamdi: Done! issue #185

demos/demo35_FD_gradient.py
amal-ghamdi: This is really a good point. I tried to capture it with some print statement and try, except clause.

tests/test_likelihood.py
amal-ghamdi: 👍

cuqi/utilities/_utilities.py
amal-ghamdi: Thank you for pointing this out. I thought about it for a bit and I think it needs more consideration, and I reverted the changes here to issue an error if the input is cuqi array. I added an issue for this #186 .

cuqi/utilities/_utilities.py
amal-ghamdi: I added more documentation

cuqi/utilities/_utilities.py
amal-ghamdi: I added more documentation

docs/conf.py
amal-ghamdi: I added an issue #187 and kept the TODO here.

amal-ghamdi:
Thank you @jakobsj for your review! I updated the PR reflecting your comments and suggestions. I appreciate it if you can take a look.

jakobsj (approved):
Everything addressed nicely, thanks!

PR #166, opened 2022-12-20
Do not allow Laplace_diff and Cauchy_diff to be defined with non-zero…

### LINKED ISSUE #165: Do not allow Cauchy diff and Laplace diff to be initialized with any other vector than zero for location
We agreed to remove location parameter entirely from Cauchy diff and Laplace diff

nabriis:
… vector

Closes #165


amal-ghamdi (approved):
LGTM!

amal-ghamdi (commented):
Thank you Nicolai, It looks great 👍. I just wanted to comment whether we need to use `np.allclose(location, 0)` instead of  `np.linalg.norm(location) > 0` in both classes to account for machine-precision-level variations.

jakobsj (approved):
Looks good to me (please consider the comment from @amal-ghamdi).

PR #167, opened 2022-12-23
Sprint16 kl projection

### LINKED ISSUE #168: Add projection (fun2par) for KLExpansion

amal-ghamdi:
Closes #168


amal-ghamdi:
Hi @jakobsj and @nabriis. I finished this KL projection pull request, but there is a general design/convention question: should we allow fun2par to be a projection or should it always be either exact or not supported. This functionality is very helpful for example in the heat1D problem where we want to look at the projected signal in the coefficient space (to get plot_ci for parameters for example). We can discuss in the stand-up.

cuqi/_cuqi_global_vars.py
```diff
@@ -0,0 +1,7 @@
+""" This module defines variables that are used throughout CUQIpy. The values of these variables are not meant to be changed by the user."""
+
+_disable_warning_msg = lambda module_name: "To disable "+\
+"warnings for a given module or library, "+\
+"you can use the method `warnings.filterwarnings`,"+\
+" e.g.: warnings.filterwarnings(action='ignore', "+\
+f"module=r'"+module_name+"')."
```
nabriis, line 7: I like the idea of having this area for warning messages etc. I would consider renaming the file to something like cuqi/_messages or something like that. Global variables should likely be in the config file.

cuqi/geometry/_geometry.py
```diff
@@ -732,7 +733,24 @@ def par2fun(self, p):
...
+        # Check that the input is of the correct shape
+        if len(funvals) != self.fun_dim:
+            raise ValueError(
+                "Input array funvals must have length {}".format(self.fun_dim))
+        
+        warnings.warn(f"fun2par for {self.__class__} is a projection on " +
+                      "the KL expansion coefficients space where only " +
+                      f"the first self.num_modes={self.num_modes} " +
+                      "coefficients are returned. " +
+                      _disable_warning_msg("cuqi.geometry")+"\n")
```
nabriis, line 745:
```suggestion
        warnings.warn(
            f"fun2par for {self.__class__} is a projection on "
            + "the KL expansion coefficients space where only "
            + f"the first self.num_modes={self.num_modes} "
            + "coefficients are returned. "
            + _disable_warning_msg("cuqi.geometry")
            + "\n"
        )

```

nabriis (approved):
After looking at it I think its OK to have this feature added. With the warning message there is no doubt that its not the full inverse.

cuqi/_cuqi_global_vars.py
amal-ghamdi: 👍

tests/test_geometry.py
```diff
@@ -361,3 +361,30 @@ def test_create_CustomKL_geometry():
...
+    # Project signal to the KL basis and back
+    p = geom.fun2par(signal)
+    assert(len(p) == num_modes)
+
+    signal_proj = geom.par2fun(p)
+    assert(len(signal_proj) == N)
+
+    # Check that the projection is accurate
+    rel_err = np.linalg.norm(signal-signal_proj)/np.linalg.norm(signal)
+    assert np.isclose(rel_err, 0.0, atol=1e-5)
```
jakobsj, line 390: Just to check my understanding, this test uses a signal that can be perfectly represented in the KLExpansion, and as such there is no loss in projecting back and forth? Good to test that, and I think it would be relevant to add at least one test of a signal that can NOT be perfectly represented - one needs to work out how to check for correctness since cannot simply compare with orignal.

cuqi/geometry/_geometry.py
```diff
@@ -732,7 +733,28 @@ def par2fun(self, p):
 
     def fun2par(self, funvals):
         """The function to parameter map used to map function values back to parameters, if available."""
```
jakobsj, line 735: Please add here that the a projection onto the coefficients is used and so the signal may not be fully represented but it will be the closest approximation in that expansion, or something like that.

cuqi/geometry/_geometry.py
```diff
@@ -732,7 +733,28 @@ def par2fun(self, p):
...
+            f"fun2par for {self.__class__} is a projection on "
+            + "the KL expansion coefficients space where only "
+            + f"the first self.num_modes={self.num_modes} "
+            + "coefficients are returned. "
+            + _disable_warning_msg("cuqi.geometry")
+            + "\n"
+        )
+
+
+        # Note, the scaling by 2*self.fun_dim is not needed if scipy.
```
jakobsj, line 751: It's good to mention the points of this comment. In general before giving technicalities, first explain what the purpose of the following code is and how the code achieves that purpose. Here specifically, we want to explain that we compute a projection onto the coefficient space etc. and why the dst and the rest of the commands produce that (maybe give a reference to where this expression comes from).

jakobsj (changes requested):
Thank you @amal-ghamdi it looks good - I've requested some minor additions to the comments and tests.

cuqi/geometry/_geometry.py
amal-ghamdi: Makes sense, I added more explanation to clarify this point.

tests/test_geometry.py
amal-ghamdi: Thank you for brining this up. Here it is not exact transformation. If the number of modes is 100 I believe it will be. The lower the number of modes the further the par2fun(fun2par(true_signal)) from the true_signal. Here I choose the number of modes to be 95, which is close to 100. and I adjusted the isclose tolerance to pass the assertion.

amal-ghamdi:
Thank you @jakobsj for your very helpful feedback! I addressed your comments, please let me know if you have further suggestions for improvements

jakobsj (approved):
Thanks for adding the extra doc, looks good!

PR #188, opened 2023-02-16
Move Data and CUQIarray classes to a separate module (array module)

### LINKED ISSUE #161: Consider creating a module for CUQIarray. `cuqi.samples.CUQIarray` seems somewhat unnatural
Our discussion today we decided to do something like this: 

Add an array module so that we have `cuqi.array.CUQIarray` and keep the samples in its module `cuqi.samples.Samples`

amal-ghamdi:
closes #161


jakobsj (approved):
Thanks @amal-ghamdi, seems a straightforward replacement so should be fine. Do we know that all instances of cuqi.samples.CUQIarray across all modules, tests, demos etc. have been replaced?

nabriis (approved):
LGTM

PR #199, opened 2023-03-01
Add Cauchy Distribution

### LINKED ISSUE #191: Add Cauchy distb

nabriis:
Closes #191

### DIFF
--- a/cuqi/distribution/_cauchy.py
+++ b/cuqi/distribution/_cauchy.py
@@ -1,0 +1,100 @@
+import numpy as np
+import scipy.stats as sps
+from cuqi.geometry import _get_identity_geometries
+from cuqi.utilities import force_ndarray
+from cuqi.distribution import Distribution
+
+class Cauchy(Distribution):
+    """ Multivariate Cauchy distribution of independent random variables.
+     
+    Each is distributed according to the PDF function
+
+    .. math::
+
+        \\frac{1}{\\pi\\gamma(1+\\frac{(x-\\mu)^2}{\\gamma^2})}
+
+    where :math:`\\mu` is the location parameter and :math:`\\gamma` is the scale parameter.
+
+    Parameters
+    ----------
+    location: float or array_like
+        Location parameter
+
+    scale: float or array_like
+        Scale parameter
+
+    Example
+    -------
+    .. code-block:: python
+
+        # % Generate a Cauchy distribution
+        import cuqi
+        location = 0
+        scale  = 1
+        x = cuqi.distribution.Cauchy(location, scale)
+        x.logpdf(0) # -1.1447298858494002
+
+    """
+
+    def __init__(self, location=None, scale=None, is_symmetric=True, **kwargs):
+        super().__init__(is_symmetric=is_symmetric, **kwargs)
+        self.location = location
+        self.scale = scale
+
+    @property
+    def location(self):
+        """ Location parameter """
+        return self._location
+    
+    @location.setter
+    def location(self, value):
+        self._location = force_ndarray(value, flatten=True)
+
+    @property
+    def scale(self):
+        """ Scale parameter """
+        return self._scale
+    
+    @scale.setter
+    def scale(self, value):
+        self._scale = force_ndarray(value, flatten=True)
+
+    def _is_out_of_bounds(self, x):
+        """ Check if x is out of bounds """
+        return np.any(self.scale <= 0)
+
+    def logpdf(self, x):
+
+        if self._is_out_of_bounds(x):
+            return -np.Inf
+        
+        return np.sum(-np.log(np.pi*self.scale*(1+((x-self.location)/self.scale)**2)))
+    
+    def cdf(self, x):
+
+        if self._is_out_of_bounds(x):
+            return -np.Inf
+        
+        return np.sum(sps.cauchy.cdf(x, loc=self.location, scale=self.scale))
+    
+    def gradient(self, x):
+
+        #Avoid complicated geometries that change the gradient.
+        if not type(self.geometry) in _get_identity_geometries():
+            raise NotImplementedError("Gradient not implemented for distribution {} with geometry {}".format(self,self.geometry))
+        
+        #Computing the gradient for conditional distribution is not supported yet    
+        if self.is_cond:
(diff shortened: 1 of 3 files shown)


cuqi/distribution/_cauchy.py
```diff
@@ -0,0 +1,100 @@
...
+
+        if self._is_out_of_bounds(x):
+            return -np.Inf
+        
+        return np.sum(-np.log(np.pi*self.scale*(1+((x-self.location)/self.scale)**2)))
+    
+    def cdf(self, x):
+
+        if self._is_out_of_bounds(x):
+            return -np.Inf
```
jakobsj, line 76: Do we have a convention for out of bounds value? Here it is -inf, later a nan vector

cuqi/distribution/_cauchy.py
```diff
@@ -0,0 +1,100 @@
...
+        if not type(self.geometry) in _get_identity_geometries():
+            raise NotImplementedError("Gradient not implemented for distribution {} with geometry {}".format(self,self.geometry))
+        
+        #Computing the gradient for conditional distribution is not supported yet    
+        if self.is_cond:
+            raise NotImplementedError(f"Gradient is not implemented for {self} with conditioning variables {self.get_conditioning_variables()}")
+        
+        # Check bounds (return nan if out of bounds)
+        if self._is_out_of_bounds(x):
+            return x*np.nan
```
jakobsj, line 92: Here nans for out of bound, previously it was -inf.

cuqi/distribution/_cauchy.py
```diff
@@ -0,0 +1,100 @@
...
+        
+        #Compute the gradient
+        return -2*(x-self.location)/(self.scale**2*(1+((x-self.location)/self.scale)**2))
```
jakobsj, line 95: Could do a few quick optimisations here, like precompute x-self.location, which is computed twice.  self.scale**2 could be stored and reused etc.

cuqi/distribution/_cauchy.py
```diff
@@ -0,0 +1,100 @@
...
+    Example
+    -------
+    .. code-block:: python
+
+        # % Generate a Cauchy distribution
+        import cuqi
+        location = 0
+        scale  = 1
+        x = cuqi.distribution.Cauchy(location, scale)
+        x.logpdf(0) # -1.1447298858494002
```
jakobsj, line 35: Add example with 2 or 3 element location and scale to demonstrate being multivariate.

jakobsj (changes requested):
Looks good, just a few minor suggestions.

cuqi/distribution/_cauchy.py
```diff
@@ -0,0 +1,100 @@
...
+        """ Scale parameter """
+        return self._scale
+    
+    @scale.setter
+    def scale(self, value):
+        self._scale = force_ndarray(value, flatten=True)
+
+    def _is_out_of_bounds(self, x):
+        """ Check if x is out of bounds """
+        return np.any(self.scale <= 0)
```
amal-ghamdi, line 64: Given that the Cauchy distribution PDF is defined on (-inf, inf) and that scale should be greater than zero, I am wondering if we need this function. The function does not actually use the value of x. Could an alternative approach be to test inside the scale setter:
```python
if np.any(value <= 0):
    #raise some error
```

tests/test_distribution.py
```diff
@@ -635,3 +636,34 @@ def test_Laplace_diff_should_not_allow_non_zero_location():
 
     with pytest.raises(ValueError):
         cuqi.distribution.Laplace_diff(lambda x: x, 1)
+
+def test_Cauchy_against_scipy():
```
amal-ghamdi, line 640: nice test 👍

amal-ghamdi (commented):
Thank you @nabriis for this PR! I added one comment about the `_is_out_of_bounds` method.

cuqi/distribution/_cauchy.py
nabriis: Yes convention is for logpdf, pdf and cdf to return -Inf and Nan otherwise.

cuqi/distribution/_cauchy.py
nabriis: Done, Did not do it for scale**2 since is not directly used twice.

cuqi/distribution/_cauchy.py
nabriis: Done

cuqi/distribution/_cauchy.py
nabriis: We do not want to raise an error since sampling algorithms may propose these kinds of values and we want the logpdf to return a valid value, but impossible.

cuqi/distribution/_cauchy.py
```diff
@@ -34,6 +34,15 @@ class Cauchy(Distribution):
         x = cuqi.distribution.Cauchy(location, scale)
         x.logpdf(0) # -1.1447298858494002
 
+    .. code-block:: python
+
+        # % Generate a multivariate Cauchy distribution
+        import cuqi
+        location = [0, 0]
+        scale  = [1, 1]
```
jakobsj, line 42: Can we make a slightly less trivial one please :)  Where location and scale are different, so that it is clear that is it not just an iid version.

cuqi/distribution/_cauchy.py
```diff
@@ -92,7 +101,8 @@ def gradient(self, x):
             return x*np.nan
         
         #Compute the gradient
-        return -2*(x-self.location)/(self.scale**2*(1+((x-self.location)/self.scale)**2))
+        x_translated = x-self.location
+        return -2*(x_translated)/(self.scale**2*(1+((x_translated)/self.scale)**2))
```
jakobsj, line 105:
```suggestion
        return -2*x_translated/(self.scale**2*(1+(x_translated/self.scale)**2))
```

jakobsj (changes requested):
Thanks!  Just a few tiny points, then it should be ready.

amal-ghamdi (approved):
LGTM!, I reviewed the recent updates, after my last review. It looks good 👍

cuqi/distribution/_cauchy.py
nabriis: Good suggestion :)

jakobsj (approved):
Thanks, looks good now!

PR #214, opened 2023-04-21
Sprint19 plot variance (generalize to more geometries)

### LINKED ISSUE #207: Implement samples.funvals.plot_variance() for Samples objects with  fenics geometries

### LINKED ISSUE #213: Expand support for computing sample statistics on function-value representation to include fenics geometries

amal-ghamdi:
closes #207


jakobsj (approved):
Quick approval, looked through but didn't check details or run. Looks to be in good shape. Merge pending more in-depth review and approval from @nabriis.

demos/tutorials/fun_vec_representation.py
```diff
@@ -0,0 +1,266 @@
+"""
```
nabriis, line 1: Thank you for this very helpful tutorial. Overall I think the flow looks good of the funvals, vector etc.

However, I would like to have some more realistic examples in this tutorial so that it is easier to reason about and understand.

My suggestion is to show-case at least these two examples in the tutorial.
1) A 2D geometry with an image taken e.g. from `cuqi.data` such as grains, threephases or astronaut. Then a mapping that e.g. does a thresholding which is clearly visible compared to the normal image. E.g. put all values below 100 to 0 or likewise.
2) A 1D geometry such as KL Expansion or StepExpansion to illustrate difference between funvals and parameters.

I believe you should be able to take the existing tutorial and put in the more realistic examples.

tests/test_samples.py
```diff
@@ -2,32 +2,127 @@
...
+                          (Samples.plot_median, True, False),
+                          (Samples.plot_variance, True, False),
+                          (Samples.plot_ci_width, True, False)])
+@pytest.mark.parametrize("geom",[
+						(cuqi.geometry.Discrete(3)),
+						(cuqi.geometry.Continuous1D(4)),
+                        (cuqi.geometry.Image2D((4,4))),
+                        (cuqi.geometry.MappedGeometry(
+                             cuqi.geometry.Image2D((3,3)),
+                             map=lambda x: x**2+2))
```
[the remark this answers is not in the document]
amal-ghamdi: Thank you Nicolai for your comment. These are very important points and I will consider them in future PRs when designing tests. I thought to keep track of these ideas, we can add it to the dev documentation. I added a small section on testing, let me know what you think, we can also postpone this documentation to be tackled in another issue. 

In regards to the tests naming convention, there seems to be a couple of suggestions out there (e.g. https://medium.com/@stefanovskyi/unit-test-naming-conventions-dd9208eadbea), may be at some point we can discuss this point. I've seen also the use of "Given-When-Then" naming scheme but without explicitly mentioning all the keywords. e.g. "test_samples_when_calling_plot_method_with_plot_par_should_raise_error".

demos/tutorials/fun_vec_representation.py
amal-ghamdi: Thank you for the suggestion! I updated the tutorial using KLExpantion and a simple 2D smile image. I think these updates clarifies the funval conversion in a much better way than the original examples in the tutorial 👍 . Let me know what you think.

nabriis (approved):
Thank you @amal-ghamdi for incorporating the suggestions  and providing further thoughts on the testing also. This LGTM now!

jakobsj (approved):
Thank you very much @amal-ghamdi for the excellent work in this PR. The vector representation now appears very elegant and consistent, which is highlighted by the comprehensive illustrations of the many ways in which one can query the shape and type of the samples contents in the tutorial. Also thanks @nabriis for the thorough review and suggestions (incl the addition of a second case) that I believe have all been incorporated by now. Happy to see this merged!

amal-ghamdi:
Thank you @nabriis and @jakobsj for your review!

I am merging it now :)

PR #249, opened 2023-06-11
Sprint20 fix detect cond dist

### LINKED ISSUE #248: Conditional distributions logd, sampling and gradient
Currently if a distribution is conditional and the user attempts to sample, evaluate logd or gradient the error thrown does not indicate the root issue, which is that the distribution is conditional.

nabriis:
Closes #248

### DIFF
--- a/cuqi/distribution/_distribution.py
+++ b/cuqi/distribution/_distribution.py
@@ -154,7 +154,7 @@ class Distribution(Density, ABC):
             kwargs = self._parse_args_add_to_kwargs(cond_vars, *args, **kwargs)
 
             # Check if all conditioning variables are specified
-            if not all([key in kwargs for key in cond_vars]):
+            if not all([key in kwargs for key in cond_vars]) or len(kwargs) < len(cond_vars)+1:
                 raise ValueError(f"{self.logd.__qualname__}: To evaluate the log density all conditioning variables must be specified. Conditioning variables are: {cond_vars}")
 
             # Extract exactly the conditioning variables from kwargs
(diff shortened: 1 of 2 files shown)


cuqi/distribution/_distribution.py
```diff
@@ -154,7 +154,7 @@ def logd(self, *args, **kwargs):
             kwargs = self._parse_args_add_to_kwargs(cond_vars, *args, **kwargs)
 
             # Check if all conditioning variables are specified
-            if not all([key in kwargs for key in cond_vars]):
+            if not all([key in kwargs for key in cond_vars]) or len(kwargs) < len(cond_vars)+1:
```
amal-ghamdi, line 157: I tried to understand the other condition  " len(kwargs) < len(cond_vars)+1"
and I think it is for cases like this below where the first clause is false and the second is true 
`x.logd( s=2)`

In this case, the error msg is "To evaluate the log density all conditioning variables must be specified. Conditioning variables are: ['s']". I thought it will be more informative if the error msg in this particular case mentions x as well?.

amal-ghamdi (commented):
Thank you @nabriis for the PR, I have a minor comment.

cuqi/distribution/_distribution.py
amal-ghamdi: where:
`x = cuqi.distribution.Gaussian(0, cov=lambda s:s)`

jakobsj: I suggest updating the comment to explain the additional condition.

jakobsj (changes requested):
Just a minor comment added to Amal's points, if you just handle this, it looks fine to me.

jakobsj (approved):

amal-ghamdi (commented):
LGTM!

amal-ghamdi (approved):

PR #257, opened 2023-08-06
Sprint20 update cont2d geom

### LINKED ISSUE #33: Add fun2par and par2fun methods to Continuous2D geometry
Currently `Continuous2D` uses default `par2fun` and `fun2par`. It needs implementation of these methods similar to what Image2D has. Not for implementation: we need to insure the reshaping is done correctly. 

```python
    def par2fun(self, pars):
        # Reshape to 2D domain (also for multiple parameter vectors). TODO: #327
        funvals = pars.reshape(self.fun_shape+(-1,)) 
        #Squeeze to return single evaluated function if only one parameter vector was given
        funvals = funvals.squeeze()
        return funvals 

    def fun2par(self, funvals):
        return funvals.ravel(order=self.order) #Maybe use reshape((self.dim,), order=self.order)

```

amal-ghamdi:
closes #33

### DIFF
--- a/cuqi/geometry/_geometry.py
+++ b/cuqi/geometry/_geometry.py
@@ -414,11 +414,22 @@ class Continuous2D(Continuous):
                 axis.set_ylabel(self.axis_labels[1])
             axis.set_aspect('equal')
 
-    def par2fun(self, par):
-        return par.reshape(self.fun_shape)
+    def par2fun(self, pars):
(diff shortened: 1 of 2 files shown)


nabriis (approved):
LGTM

amal-ghamdi:
Thank you @jakobsj and @nabriis for your review!

PR #264, opened 2023-08-18
Sprint20 fix model output type

amal-ghamdi:


amal-ghamdi:
closed because wrong base branch is used. Another one is opened #265

PR #265, opened 2023-08-19
fix output of model

amal-ghamdi:
closes https://github.com/CUQI-DTU/CUQIpy-FEniCS/issues/35

The related CUQIpy-FEniCS issue needed to be fixed in the CUQIpy repo

### DIFF
--- a/cuqi/model/_model.py
+++ b/cuqi/model/_model.py
@@ -175,7 +175,7 @@ class Model(object):
         else:
             return x
 
-    def _output2par(self, out, geometry, to_CUQIarray=False):
+    def _output2par(self, out, geometry, to_CUQIarray=False, is_par=False):
         """ Converts output of :class:~`cuqi.model.Model` operators (e.g. _forward_func, _adjoint_func, _gradient_func) to parameters using the appropriate geometry.
 
         Parameters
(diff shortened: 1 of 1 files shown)


cuqi/model/_model.py
```diff
@@ -148,57 +148,82 @@ def domain_dim(self):
...
         ndarray or cuqi.array.CUQIarray
-            The input value represented as a function.
+            `x` represented as a function.
         """
-        if type(x) is CUQIarray and not isinstance(x.geometry, _DefaultGeometry1D):
-            return x.funvals
+        # Convert to function representation
+        # if x is CUQIarray and geometry are consistent, we obtain funvals
+        # directly
+        if isinstance(x, CUQIarray) and  x.geometry == geometry:
```
nabriis, line 178: Just to be sure. Can we compare the compute time when input is CUQIarray for Models now? Is the geometry check causing a lot of overhead?

nabriis (commented):
Looks good to me. I raise one concern relating to the potential extra overhead introduced when input is CUQIarray to Model forward evaluation

jakobsj (approved):
Looks good, no additional comments, but good point raised by @nabr

amal-ghamdi:
Thank you @nabriis and @jakobsj for your review. It is a good point about the overhead. I documented the overhead for a couple of examples in the attached file. It seems alright, but let me know what you think.


[overhead_geom_check.pdf](https://github.com/CUQI-DTU/CUQIpy/files/12494563/overhead_geom_check.pdf)

nabriis (approved):

PR #275, opened 2023-09-21
Add cookie phantom and update Deconvolution2D to use cuqi.data

### LINKED ISSUE #242: Add phantom="cookie"

nabriis:
Closes #242


cuqi/testproblem/_testproblem.py
```diff
@@ -1156,11 +1156,18 @@ class Deconvolution2D(BayesianProblem):
...
+        | If ndarray it should be a 2D array representing an image.
+        | The image will automatically be resized to fit the problem size.
+        | If string it should be any 2D phantom defined in cuqi.data.
+        | The string is lowercased and any hyphens are replaced 
+        | with underscores to match a method name in cuqi.data.
+        | Examples:
         | 'astronaut' - a photo of an astronaut.
         | 'camera' - a photo of a man with a camera.
         | 'cat' - a photo of a cat.
+        | 'cookie' - cartoon cookie.
```
amal-ghamdi, line 1170: I have few comments here for your consideration.

1- Is phantom limited to square phantoms here? if so maybe this can be mentioned in the documentation when talking about the possibility of passing 2D array. If not, then do we need to update the property dim to be tuple? to be reasonable for non-square images.

2- The module cuqi.data has the following components  [satellite, astronaut, camera, cat, cookie, grains, shepp_logan, threephases, p_power, rgb2gray, imresize]. I think it will be nice to give the user a way to easily and immediately distinguish which are 1D phantoms, 2D phantoms and neither of these methods. Possible way to achieve that is to create 2 methods: `get_1D_phantom_names()` and `get_2D_phantom_names()` so the user can have the list (that could grow in the future) of phantoms. This also can be written in the documentation here: "To get the list of all 2D phantom, call cuqi.data.get_1D_phantom_names()"

3- if (2) above is addressed, maybe there is no need to mention "The string is lowercased and any hyphens are replaced with underscores to match a method name in cuqi.data." as it seems a bit technical and the user will have access to the phantom name list . Might say, phantom name is case insensitive.

amal-ghamdi (commented):
Thank you @nabriis for this addition! It looks great, I only have few comments for your consideration.

One additional point, if there are already tests for the 2D phantoms, maybe the test could be expanded to include the cookie phantom (by adding new test parameter).

cuqi/testproblem/_testproblem.py
nabriis: Added issue #278. Updated wording to reflect image should be square

cuqi/data/_data.py
```diff
@@ -315,7 +316,40 @@ def p_power(size=128, relnz=0.3, p=2, seed=1): #relnz=0.65, p=2.3
     
     return x
 
+def cookie(size=128, grayscale=True):
+    """ Cartoon-style image of a cookie.
 
+    Parameters
+    ----------
+    size : int
+        Size of the image to generate. Image is square with sides of length size.
```
jakobsj, line 325: COuld you please state somewhere around here
- that the image is generated by loading it in from a PNG file
- what the original size of the image in the PNG is (so that if one wants to avoid resampling then one knows what size to use
- consider if the default size should be the non-resampled size

cuqi/data/_data.py
```diff
@@ -315,7 +316,40 @@ def p_power(size=128, relnz=0.3, p=2, seed=1): #relnz=0.65, p=2.3
...
+    stream = pkg_resources.resource_stream(__name__, 'cookie.png')
+    cookie = plt.imread(stream)
+
+    # Convert to rgb
+    cookie = cookie[..., :3]
+
+    # Convert to grayscale
+    if grayscale:
+        cookie = rgb2gray(cookie)
+        cookie[cookie < 0.05] = 0 # Make background completely black
```
jakobsj, line 347: I like this but feel it should be stated in the doc string or similar so that the user will not be surprised that the loaded in image does not agree 100% with the PNG file.

cuqi/data/_data.py
```diff
@@ -315,7 +316,40 @@ def p_power(size=128, relnz=0.3, p=2, seed=1): #relnz=0.65, p=2.3
...
+
+    # Resize
+    cookie = imresize(cookie, size)
+
+    return cookie
 
 def rgb2gray(img):
     return img @ np.array([0.2125, 0.7154, 0.0721])
```
jakobsj, line 355: Could you add a reference for where are these conversion numbers taken from? If I search for rgb2gray with python or matlab it seems different numbers ( 0.2989 * R + 0.5870 * G + 0.1140 * B ) are used as standard:
https://stackoverflow.com/questions/12201577/how-can-i-convert-an-rgb-image-into-grayscale-in-python
https://se.mathworks.com/help/matlab/ref/rgb2gray.html

jakobsj (changes requested):
@nabriis Thanks for adding the nice cookie phantom!  OVerall the PR looks good and straightforward and I just added a few minor comments.

cuqi/data/_data.py
nabriis: Done

cuqi/data/_data.py
nabriis: Good point. Added

cuqi/testproblem/_testproblem.py
amal-ghamdi: Thank you @nabriis for addressing this.

amal-ghamdi (approved):
LGTM!

jakobsj (approved):
Looks great, thanks for adding the additional documentation, in particular explaining the choices of weights for conversion from rgb to gray, I didn't know about that!

PR #292, opened 2023-10-11
checkpoint added

babakmaboudi:
This is the rework of the CUQIpy samplers:

The parent calss "Sampler_new" is a generic class for which it handles samples. Ideally it should also handle sampler management but at this point it doesn't do anything.

The child class "ProposalBasedSampler" handles the proposal and acceptance/rejection history. In ideal case pCN should pass its prior to the __init__  of ProposalBasedSampler as the proposal strategy. But I couldn't make that work.

The grandchild class MH or pCN is in charge of proposing and accepting/rejecting a single sample. ProposalBasedSampler is in charge of tracking the history. However, I didn't mange to make pCN work.

Let me know if you  have questions.
Babak

### DIFF
--- a/cuqi/sampler/__init__.py
+++ b/cuqi/sampler/__init__.py
@@ -6,6 +6,6 @@ from ._gibbs import Gibbs
 from ._hmc import NUTS
 from ._langevin_algorithm import ULA, MALA
 from ._laplace_approximation import UGLA
-from ._mh import MH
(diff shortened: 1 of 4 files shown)


nabriis:
Thank you very much for adding the checkpoint PR @babakmaboudi! Super excited to get this sampler rework into CUQIpy! It has a lot of potential.

PR #302, opened 2023-10-15
change get_components to regular method

### LINKED ISSUE #47: Tab completion of get_components
I am not too happy with our get_components method of BayesianProblem. One major issue is that there is no tab completion on the get_components method because its a class_method and it needs the init parameters from the class.

Perhaps we can actually use the init of the BayesianProblem but add a property like

``` Python
A, y_obs, probInfo = Deconvolution1D.get_components(dim=50) # Old
A, y_obs, probInfo = Deconvolution1D(dim=50).model_data_info # New
```

This matches the python programming style and it also makes it more clear what the property does due to the name scheme. Finally it also opens up for other properties that could be accessed like that. For example

``` Python
A  = Deconvolution1D(dim=50).model # Only model
y_obs  = Deconvolution1D(dim=50).data # Only data
```
etc

Thoughts?

chaozg:
This PR fixed #47 by changing `get_components` from a class method to a regular/instance method.

The following snippet is used to check if the returned components are as expected:
```Python
# %%
import cuqi
from cuqi.testproblem import Deconvolution1D

# %% setup problems
dim = [32, 128]
phantom = ["pc","bumps"]
noise_type = ["Gaussian","ScaledGaussian"]
noise_std = [0.05, 0.0001]

# Test problems
prob0 = Deconvolution1D(
    dim = dim[0],
    phantom=phantom[0],
    noise_type=noise_type[0],
    noise_std = noise_std[0]
)
prob1 = Deconvolution1D(
    dim = dim[1],
    phantom=phantom[1],
    noise_type=noise_type[1],
    noise_std = noise_std[1]
)

# %% use get_components to extract information
A0, y_data0, probInfo0 = prob0.get_components()
A1, y_data1, probInfo1 = prob1.get_components()

# check if the extraced y_datax are as expected
plt.figure()
y_data0.plot()
plt.figure()
y_data1.plot()
```
and the plots are like:
image
and 
image

### DIFF
--- a/demos/howtos/Defining_Posterior.py
+++ b/demos/howtos/Defining_Posterior.py
@@ -70,11 +70,8 @@ posterior.logd(np.ones(A.domain_dim))
 #    \end{align*}
 
 # Both observations come from the same unknown x
-A, y_obs, _ = cuqi.testproblem.Deconvolution1D.get_components()
-B, d_obs, _ = cuqi.testproblem.Deconvolution1D.get_components(
-    PSF="Defocus",
-    noise_std=0.02
-)
+A, y_obs, _ = cuqi.testproblem.Deconvolution1D().get_components()
+B, d_obs, _ = cuqi.testproblem.Deconvolution1D(PSF="Defocus",noise_std=0.02).get_components()
 
 # %%
 # Then consider the following Bayesian model
@@ -127,11 +124,8 @@ posterior2.logd(np.ones(A.domain_dim))
 # where :math:`C` is a nonlinear function.
 
 # Same x for all 3 observations
-A, y_obs, _ = cuqi.testproblem.Deconvolution1D.get_components()
-B, d_obs, _ = cuqi.testproblem.Deconvolution1D.get_components(
-    PSF="Defocus",
-    noise_std=0.02
-)
+A, y_obs, _ = cuqi.testproblem.Deconvolution1D().get_components()
+B, d_obs, _ = cuqi.testproblem.Deconvolution1D(PSF="Defocus", noise_std=0.02).get_components()
 C = cuqi.model.Model(lambda x: np.linalg.norm(x)**2, 1, A.domain_dim)
 b_obs = 16
 
--- a/demos/howtos/Defining_Posterior.py
+++ b/demos/howtos/Defining_Posterior.py
@@ -71,7 +71,7 @@ posterior.logd(np.ones(A.domain_dim))
 
 # Both observations come from the same unknown x
 A, y_obs, _ = cuqi.testproblem.Deconvolution1D().get_components()
-B, d_obs, _ = cuqi.testproblem.Deconvolution1D(PSF="Defocus",noise_std=0.02).get_components()
+B, d_obs, _ = cuqi.testproblem.Deconvolution1D(PSF="Defocus", noise_std=0.02).get_components()
 
 # %%
 # Then consider the following Bayesian model
--- a/demos/howtos/Defining_Posterior.py
+++ b/demos/howtos/Defining_Posterior.py
@@ -23,7 +23,7 @@ import numpy as np
 #
 # See :class:`~cuqi.testproblem.Deconvolution1D` for more details.
 
-A, y_obs, _ = cuqi.testproblem.Deconvolution1D.get_components()
+A, y_obs, _ = cuqi.testproblem.Deconvolution1D().get_components()
 
 # %%
 # Then consider the following Bayesian model
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ from cuqi.distribution import Gaussian, LMRF, Gamma
 from cuqi.problem import BayesianProblem
 
 # Step 1: Set up forward model and data, y = Ax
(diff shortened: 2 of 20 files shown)


chaozg:
The failure is due to a check at `/tests/test_bayesian_inversion.py`

```Python
def test_Bayesian_inversion_hierarchical(TP_type: BayesianProblem, phantom: str, priors: Dict[str, Density], Ns: int):
    # Load model + data from testproblem library
    A, y_data, probInfo = TP_type.get_components(dim=priors[0].dim, phantom=phantom)
```
I think I need some advice on how to solve this problem.

nabriis (changes requested):
Thank you @chaozg. The current changes look great and the actual change to the BayesianProblem class is exactly as I had hoped!

I would advice using a search tool like the search function inside visual studio code to find all cases of "get_components" to ensure to did not miss any. For example there should be some inside the docs/ folder also.

In addition, I think we should briefly discuss at the stand up what the best new syntax is, e.g.:
```python
A, y_obs, info = Deconvolution1D(dim=128).get_components() # Current with this case
A, y_obs, info = Deconvolution1D(dim=128).components # Perhaps this more concice syntax?
```

Finally, considering that we call it "components" of the BayesianProblem. It may be worth it to ensure we actually include all the components from the problem. We could do this by adding more things to the ProblemInfo class. This could e.g. be the likelihood and prior also. I note this could be added as a new issue and PR as to not polute this one too much

chaozg:
> I would advice using a search tool like the search function inside visual studio code to find all cases of "get_components" to ensure to did not miss any. For example there should be some inside the docs/ folder also.

@nabriis Thanks for the hint, and yes I just did as you suggested and confirm that all files including `get_components` run well now.
Meanwhile, I want to make an off-topic note that `demos/ExploreTestProblems` mentions `ParBeamCT2D`, which seems not available in CUQIpy's current release:
```Python
elif testprob == 5:
    model, data, probInfo = ParBeamCT2D(im_size=(50,50),det_count=60, proj_type="line").get_components() #cuda
```
Anyhow, I also made changes according here, i.e., moving arguments `im_size=(50,50),det_count=60, proj_type="line"` from `get_components` to the class constructor.

demos/howtos/Defining_Posterior.py
```diff
@@ -70,11 +70,8 @@
 #    \end{align*}
 
 # Both observations come from the same unknown x
-A, y_obs, _ = cuqi.testproblem.Deconvolution1D.get_components()
-B, d_obs, _ = cuqi.testproblem.Deconvolution1D.get_components(
-    PSF="Defocus",
-    noise_std=0.02
-)
+A, y_obs, _ = cuqi.testproblem.Deconvolution1D().get_components()
+B, d_obs, _ = cuqi.testproblem.Deconvolution1D(PSF="Defocus",noise_std=0.02).get_components()
```
amal-ghamdi, line 74:
```suggestion
B, d_obs, _ = cuqi.testproblem.Deconvolution1D(PSF="Defocus", noise_std=0.02).get_components()
```

amal-ghamdi (commented):
Thank you Charlie for this update! The PR looks good 👍 I just had one minor suggested change. Also thank you for creating issue #https://github.com/CUQI-DTU/CUQIpy-CIL/issues/28 , would you please create a similar one for CUQIpy-FEniCS. Thank you.

amal-ghamdi:
> ```python
> ParBeamCT2D
> ```

That's a good catch! I think this test problem was moved to CIL `cuqipy_cil.testproblem.ParallelBeam2D`, @jakobsj and @nabriis might be able to confirm if this is the case 😊

jakobsj:
I parked the potential renaming of get_components() as a separate issue here https://github.com/CUQI-DTU/CUQIpy/issues/303

nabriis:
> > I would advice using a search tool like the search function inside visual studio code to find all cases of "get_components" to ensure to did not miss any. For example there should be some inside the docs/ folder also.
> 
> @nabriis Thanks for the hint, and yes I just did as you suggested and confirm that all files including `get_components` run well now. Meanwhile, I want to make an off-topic note that `demos/ExploreTestProblems` mentions `ParBeamCT2D`, which seems not available in CUQIpy's current release:
> 
> ```python
> elif testprob == 5:
>     model, data, probInfo = ParBeamCT2D(im_size=(50,50),det_count=60, proj_type="line").get_components() #cuda
> ```
> 
> Anyhow, I also made changes according here, i.e., moving arguments `im_size=(50,50),det_count=60, proj_type="line"` from `get_components` to the class constructor.

Hello @chaozg. Thanks for updating that. I believe there must be other areas where get_components is used. Two that I noticed are in the index.rst file for online docs and in the example docstring for BayesianProblem.  Did you use the search functionality to find all uses of get_components?

nabriis (commented):
See comment above. Seems more areas need to be updated.

nabriis (approved):
Thank you @chaozg. The PR LGTM now! 🎉 

Thank you for sticking through the review process. I think this change will make the testproblems much easier to use for the users! Very important.

amal-ghamdi (approved):
Thank you @chaozg for the updates! LGTM

PR #309, opened 2023-11-08
fix get_indirect_variables as suggested in issue #289

### LINKED ISSUE #289: Fix CUQIpy conditioning variables being extracted from private variables
Update utilities to this:

```python
def get_indirect_variables(dist):
    attributes = []
    for attribute in dist.get_mutable_variables():
        value = getattr(dist, attribute)
        if callable(value):
            keys = get_non_default_args(value)
            for key in keys:
                if key not in attributes: #Ensure we did not already find this key
                    attributes.append(key)
    return attributes 
```

chaozg:
updated `utilities/_utilities/get_indirect_variables` as suggested in #289 .

Amal: I am adding (closes #289) to link the issue with this PR :)

### DIFF
--- a/cuqi/utilities/_utilities.py
+++ b/cuqi/utilities/_utilities.py
@@ -62,7 +62,8 @@ def get_direct_attributes(dist):
 
 def get_indirect_variables(dist):
     attributes = []
-    for _, value in vars(dist).items():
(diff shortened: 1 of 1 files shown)


amal-ghamdi (approved):
LGTM!

PR #327, opened 2023-12-07
Allow lazy import of arviz

nabriis:
Allows lazy import of arviz. This means user with install issues can run `pip install cuqipy --no-deps` and likely get most of the functionality to work


amal-ghamdi (approved):
LGTM! great timing :)

PR #340, opened 2023-12-21
Fix bug in distribution gallery

### LINKED ISSUE #339: Fix bug in BivariateGaussian from DistributionGallery
The following snippet produces an error as the constructed covariance matrix is not symmetric
```python
import cuqi
from cuqi.distribution import DistributionGallery
dist = DistributionGallery("BivariateGaussian")
```

Relevant problematic lines from `cuqi/distribution/_custom.py` are:
```python
  dim = 2
  mu = np.zeros(dim)
  sigma = np.linspace(0.5, 1, dim)
  R = np.array([[1.0, .9 ],[.9, 1]])
  dist = Gaussian(mu, (sigma**2)*R)
```

The above code gives covariance `(sigma**2)*R` as:
```python
array([[0.25 , 0.9  ],
       [0.225, 1.   ]])
```
and it's not symmetric.

chaozg:
fix #339

### DIFF
--- a/cuqi/distribution/_custom.py
+++ b/cuqi/distribution/_custom.py
@@ -10,7 +10,7 @@ class UserDefinedDistribution(Distribution):
     ------------
     logpdf_func: Function evaluating log probability density function. Callable.
     gradient_func: Function evaluating the gradient of the logpdf. Callable.
-    logpdf_func: Function evaluating log probability density function. Callable.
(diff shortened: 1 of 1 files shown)


amal-ghamdi (approved):
LGTM!

Good catch, @chaozg!

jakobsj (approved):
Nice!

Issue #331, opened 2023-12-08
Multivariate Gamma returns a single value


saveriDTU:
**Description**

Only one value is returned when sampling from a multivariate Gamma distribution.

**Example to reproduce**

import cuqi
from cuqi.distribution import Gamma
d = Gamma(1, 1e-4, geometry=50)
ds = d.sample()
ds

Out[1]: 
CUQIarray: NumPy array wrapped with geometry.
---------------------------------------------

Geometry:
 _DefaultGeometry1D(50,)

Parameters:
 True

Array:
CUQIarray(680.24372138)

PR #358, opened 2024-02-02
Make Gamma distribution multivariate

### LINKED ISSUE #330: Gamma distribution multivariate
**Description**

Describe the issue and what the expected result should be

**Example to reproduce**

Provide a code example to reproduce the bug

```python
import numpy as np
from cuqi.distribution import Gamma
test = Gamma(np.zeros(10), np.ones(10))
test.sample()
```

nabriis:
Closes #330 #331

### DIFF
--- a/cuqi/distribution/_gamma.py
+++ b/cuqi/distribution/_gamma.py
@@ -1,9 +1,43 @@
 import numpy as np
+import scipy.stats as sps
 from scipy.special import loggamma, gammainc
 from cuqi.distribution import Distribution
 
 class Gamma(Distribution):
+    """
+    Represents a multivariate Gamma distribution characterized by shape and rate parameters of independent random variables x_i. Each is distributed according to the PDF function
+    
+    f(x; shape, rate) = rate^shape * x^(shape-1) * exp(-rate * x) / Gamma(shape)
 
+    where `shape` and `rate` are the parameters of the distribution, and Gamma is the Gamma function.
+
+    Parameters
+    ----------
+    shape : float or array_like, optional
+        The shape parameter of the Gamma distribution. Must be positive.
+
+    rate : float or array_like, optional
+        The rate parameter of the Gamma distribution. Must be positive. The inverse of the scale parameter.
+
+    Examples
+    --------
+    .. code-block:: python
+
+        import numpy as np
+        import cuqi
+        import matplotlib.pyplot as plt
+
+        # Create a Gamma distribution instance
+        shape = 1
+        rate = 1e-4
+        gamma_dist = cuqi.distribution.Gamma(shape=shape, rate=rate)
+
+        # Generate samples
+        samples = gamma_dist.sample(10000)
+
+        # Plot histogram of samples
+        samples.hist_chain(0, bins=70)
+    """
     def __init__(self, shape=None, rate=None, is_symmetric=False, **kwargs):
         # Init from abstract distribution class
         super().__init__(is_symmetric=is_symmetric,**kwargs) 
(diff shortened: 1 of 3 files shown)


cuqi/distribution/_gamma.py
```diff
@@ -1,37 +1,79 @@
...
-from scipy.special import loggamma, gammainc
+import scipy.stats as sps
 from cuqi.distribution import Distribution
+from cuqi.utilities import force_ndarray
 
 class Gamma(Distribution):
+    """
+    Represents a multivariate Gamma distribution characterized by shape and rate parameters of independent random variables x_i. Each is distributed according to the PDF function
+    
+    f(x; shape, rate) = rate^shape * x^(shape-1) * exp(-rate * x) / Gamma(shape)
```
amal-ghamdi, line 10: following @jakobsj 's suggestion of adding index i, the index i can be added to the shape and rate as well

cuqi/distribution/_gamma.py
```diff
@@ -1,37 +1,79 @@
...
 
-    def pdf(self, x):
-        # sps.gamma.pdf(x, a=self.shape, loc=0, scale=self.scale)
-        # (self.rate**self.shape)/(gamma(self.shape)) * (x**(self.shape-1)*np.exp(-self.rate*x))
-        return np.exp(self.logpdf(x))
-
     def logpdf(self, x):
-        # sps.gamma.logpdf(x, a=self.shape, loc=0, scale=self.scale)
-        return (self.shape*np.log(self.rate)-loggamma(self.shape)) + ((self.shape-1)*np.log(x) - self.rate*x)
+        return np.sum(sps.gamma.logpdf(x, a=self.shape, loc=0, scale=self.scale))
```
amal-ghamdi, line 69: I see sps.gamma support loc parameter. Do we want to support that as well? as initializer parameter?

cuqi/distribution/_gamma.py
```diff
@@ -1,37 +1,79 @@
 import numpy as np
-from scipy.special import loggamma, gammainc
+import scipy.stats as sps
 from cuqi.distribution import Distribution
+from cuqi.utilities import force_ndarray
 
 class Gamma(Distribution):
+    """
```
amal-ghamdi, line 7: I suggest adding a unit test for the multivariate logpdf/sample/cdf.

amal-ghamdi (changes requested):
Thank you @nabriis for this fix! I have only 3 comments for your consideration.

cuqi/distribution/_gamma.py
nabriis: If requested in the future we can add an issue for it IMO.

cuqi/distribution/_gamma.py
nabriis: Yes

cuqi/distribution/_gamma.py
```diff
@@ -1,37 +1,79 @@
...
+    .. code-block:: python
+
+        import numpy as np
+        import cuqi
+        import matplotlib.pyplot as plt
+
+        # Create a Gamma distribution instance
+        shape = 1
+        rate = 1e-4
+        gamma_dist = cuqi.distribution.Gamma(shape=shape, rate=rate)
```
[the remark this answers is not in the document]
nabriis: Yes

cuqi/distribution/_gamma.py
```diff
@@ -1,37 +1,79 @@
...
 
+    where `shape` and `rate` are the parameters of the distribution, and Gamma is the Gamma function.
+
+    Parameters
+    ----------
+    shape : float or array_like, optional
+        The shape parameter of the Gamma distribution. Must be positive.
+
+    rate : float or array_like, optional
+        The rate parameter of the Gamma distribution. Must be positive. The inverse of the scale parameter.
```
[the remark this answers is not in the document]
nabriis: Added it as description for the attribute

amal-ghamdi (approved):
Thank you for the update. LGTM!

nabriis:
> Thank you for adding this, it seems fairly straightforward so not many comments. I note from the error message that this PR stops short of allowing multivariate Gamma with the conjugate sampler. How big a job would that be?

Unsure how much effort it would be to support multivariate in conjugate. Likely just need to extend the equations to multivariate case, where now they assume scalar.

cuqi/distribution/_gamma.py
```diff
@@ -27,16 +31,35 @@ class Gamma(Distribution):
         import cuqi
         import matplotlib.pyplot as plt
 
-        # Create a Gamma distribution instance
+        # Create a multivariate Gamma distribution with the same shape and rate parameters
```
jakobsj, line 34:
```suggestion
        # Create a 10-dimensional multivariate Gamma distribution with the same shape and rate parameters
```

jakobsj (approved):
OK. Please park the conjugacy point as a new issue, refer to it here, consider the small suggestion to clarify the dimension in the example, and then go ahead and merge.

PR #361, opened 2024-02-09
Distributions: Disable conditioning on mutable variables if not conditioning variable

### LINKED ISSUE #323: Disable conditioning on mutable variables for distributions
Currently it is possible to condition on mutable variables (e.g. mean, cov etc.) of distributons even if they are not conditioning varibles. For example

```python
x = Gaussian(0, 1)
x(mean=1) # Changes mean to 1
```
or
```python
x = Gaussian(0, lambda s: s)
x(cov=1) # changes cov to 1, and removed s dependancy
```

We have discussed to **not** allow this behavior, and only allow conditioning on variables that the distribution is conditioned on.

This means the above case is not allowed but these are still allowed

```python
x = Gaussian(0, lambda s: s)
x(s=1) # Changes covariance to 1
```` 

```python
x = Gaussian(geometry=1) # mean and cov not set, so they are conditioning variables.
x(mean=1, cov=1) # mean and cov set via conditioning.
````

nabriis:
Closes #323 (Updated issue text)

The updates to the unit tests reflect our choice of not allowing condiotining on mutable variables if they are in fact not conditioning variables.

### DIFF
--- a/cuqi/distribution/_distribution.py
+++ b/cuqi/distribution/_distribution.py
@@ -284,6 +284,12 @@ class Distribution(Density, ABC):
         new_dist = self._make_copy() #New cuqi distribution conditioned on the kwargs
         processed_kwargs = set() # Keep track of processed (unique) elements in kwargs
 
+        # Check if kwargs contain any mutable variables that are not conditioning variables
+        # If so we raise an error since these are not allowed to be specified.
+        for kw_key in kwargs.keys():
+            if kw_key in mutable_vars and kw_key not in cond_vars:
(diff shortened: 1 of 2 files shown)


nabriis:
Hey @jakobsj. Thanks for the comments. I can understand why it would be unclear. I updated the issue and PR description. Hope this provides enough information now :)

amal-ghamdi (approved):
LGTM!

jakobsj (approved):
Thank you for updating the description, much clearer now :)   It looks good to me.  Nice to see that the unit tests caught the change :)

PR #363, opened 2024-02-16
Sprint27 beta grad

### LINKED ISSUE #159: When Beta distribution is used as a likelihood (e.g. mean is a cuqi model), it should either handle the gradient by applying chain rule or generate an error that it does not handle the gradient.

amal-ghamdi:
closes #159 

There is an error raised already in the case described which state: "NotImplementedError: Gradient is not implemented for CUQI Beta. Conditioning variables ['x']. with conditioning variables ['x']". However, computing the gradient for Beta likelihood used to (before this PR) fail for another reason: Beta._gradient() takes 2 positional arguments but 3 were given. I fixed the latter error by updating the gradient argument to be consistent with, e.g., Gaussian gradient. I also added a test. Now the first error is raised when attempting to compute the gradient for Beta likelihood.

### DIFF
--- a/cuqi/distribution/_beta.py
+++ b/cuqi/distribution/_beta.py
@@ -69,7 +69,7 @@ class Beta(Distribution):
     def _sample(self, N=1, rng=None):
         return sps.beta.rvs(a=self.alpha, b=self.beta, size=(N,self.dim), random_state=rng).T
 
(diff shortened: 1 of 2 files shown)


jakobsj (approved):
LGTM

nabriis (approved):
LGTM

PR #376, opened 2024-03-06
Sampler rework: refactor CWMH sampler

### LINKED ISSUE #371: Subclass the CWMH samplers from the new Sampler class

amal-ghamdi:
closes #371 

Things to consider for the review:
- The code is mainly copied from the original `CWMH`
- Modification to remove all the outer sampling loops and just keep the step and tune parts
- I needed to introduce additional scale parameter to keep track of old scale (to enable matching the old and new sampler)
- I modified the class documentation a bit, including stating that scale can be an array in this CWMH case. And updated the example code in the class documentation
-  I modified the tests method `assert_true_if_sampling_is_equivalent` and `assert_true_if_warmup_is_equivalent` to accept additional inputs: `old_idx` and `new_idx`. Which I use for example to remove first sample in the new sampler which is the initial point, that is not included in the samples of the old sampler. I use it also to remove the last sample in the old sampler because it is repeated (same as second to last sample)

Question:
- [x] I am not sure about the dim property? it seems like it was remover from the `NewSampler`. Should we still support it?(I removed dim from CWMH as discussed)

### DIFF
--- a/cuqi/experimental/mcmc/_cwmh.py
+++ b/cuqi/experimental/mcmc/_cwmh.py
@@ -124,6 +124,8 @@ class CWMHNew(ProposalBasedSamplerNew):
             raise ValueError(fail_msg)
 
     def step(self): #CWMH_new
+        """ Perform one step of the sampler by transitioning the current point
+        to a new point according to the sampler's transition kernel. """
         # Propose state x_i_star used to update
         # each component of x_t step by step
         x_t = self.current_point.copy()
@@ -166,6 +168,8 @@ class CWMHNew(ProposalBasedSamplerNew):
         return acc
 
     def tune(self, skip_len, update_count):
+        """ Tune the parameters of the sampler. This method is called after each
+        step of the warmup phase. """
         idx = update_count
         star_acc = 0.21/self.dim + 0.23
         hat_acc = np.mean(self._acc[idx*skip_len:(idx+1)*skip_len], axis=0)
--- a/cuqi/experimental/mcmc/_cwmh.py
+++ b/cuqi/experimental/mcmc/_cwmh.py
@@ -123,7 +123,7 @@ class CWMHNew(ProposalBasedSamplerNew):
         else:
             raise ValueError(fail_msg)
 
-    def step(self): #CWMH_new
+    def step(self):
         """ Perform one step of the sampler by transitioning the current point
         to a new point according to the sampler's transition kernel. """
         # Propose state x_i_star used to update
--- a/cuqi/experimental/mcmc/_cwmh.py
+++ b/cuqi/experimental/mcmc/_cwmh.py
@@ -1,0 +1,149 @@
+import numpy as np
+import cuqi
+from cuqi.mcmc import ProposalBasedSamplerNew
+from cuqi.array import CUQIarray
+
+class CWMH_new(ProposalBasedSamplerNew):
+    """Component-wise Metropolis Hastings sampler.
+
+    Allows sampling of a target distribution by a component-wise random-walk sampling of a proposal distribution along with an accept/reject step.
+
+    Parameters
+    ----------
+
+    target : `cuqi.distribution.Distribution` or lambda function
+        The target distribution to sample. Custom logpdfs are supported by using a :class:`cuqi.distribution.UserDefinedDistribution`.
+    
+    proposal : `cuqi.distribution.Distribution` or callable method
+        The proposal to sample from. If a callable method it should provide a single independent sample from proposal distribution. Defaults to a Gaussian proposal.  *Optional*.
+
+    scale : float
+        Scale parameter used to define correlation between previous and proposed sample in random-walk.  *Optional*.
+
+    x0 : ndarray
+        Initial parameters. *Optional*
+
+    dim : int
+        Dimension of parameter space. Required if target and proposal are callable functions. *Optional*.
+
+    callback : callable, *Optional*
+        If set this function will be called after every sample.
+        The signature of the callback function is `callback(sample, sample_index)`,
+        where `sample` is the current sample and `sample_index` is the index of the sample.
+        An example is shown in demos/demo31_callback.py.
+
+    Example
+    -------
+    .. code-block:: python
+
+        # Parameters
+        dim = 5 # Dimension of distribution
+        mu = np.arange(dim) # Mean of Gaussian
+        std = 1 # standard deviation of Gaussian
+
+        # Logpdf function
+        logpdf_func = lambda x: -1/(std**2)*np.sum((x-mu)**2)
+
+        # Define distribution from logpdf as UserDefinedDistribution (sample and gradients also supported as inputs to UserDefinedDistribution)
+        target = cuqi.distribution.UserDefinedDistribution(dim=dim, logpdf_func=logpdf_func)
+
+        # Set up sampler
+        sampler = cuqi.sampler.CWMH(target, scale=1)
+
+        # Sample
+        samples = sampler.sample(2000)
+
+    """
+    def __init__(self, target: cuqi.density.Density, proposal=None, scale=1,
+                 initial_point=None, dim=None, **kwargs):
+        super().__init__(target, proposal=proposal, scale=scale,
+                         x0=initial_point, dim=dim, **kwargs)
+
+    @ProposalBasedSamplerNew.proposal.setter 
+    def proposal(self, value):
+        fail_msg = "Proposal should be either None, "+\
+            f"{self.distribution.Distribution.__class__.__name__} "+\
+            "conditioned only on 'location' and 'scale', lambda function, "+\
+            f"or {self.distribution.Normal.__class__.__name__} conditioned "+\
+            "only on 'mean' and 'std'"
+
+        if value is None:
+            self._proposal = cuqi.distribution.Normal(mean = lambda location:location,std = lambda scale:scale, geometry=self.dim)
+
+        elif isinstance(value, cuqi.distribution.Distribution) and sorted(value.get_conditioning_variables())==['location','scale']:
+            self._proposal = value
+
+        elif isinstance(value, cuqi.distribution.Normal) and sorted(value.get_conditioning_variables())==['mean','std']:
+            self._proposal = value(mean = lambda location:location, std = lambda scale:scale)
+
+        elif not isinstance(value, cuqi.distribution.Distribution) and callable(value):
+            self._proposal = value
+
+        else:
+            raise ValueError(fail_msg)
+        
+        #self._proposal.geometry = self.target.geometry
+
+    def step(self, x_t, target_eval_t): #CWMH_new
+        # Propose state x_i_star used to update x_t
+        # each component of x_t step by step 
+        if isinstance(self.proposal,cuqi.distribution.Distribution):
+            x_i_star = self.proposal(location= x_t, scale = self.scale).sample()
+        else:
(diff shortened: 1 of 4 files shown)


cuqi/experimental/mcmc/_cwmh.py
```diff
@@ -0,0 +1,207 @@
...
+    scale : float or ndarray
+        Scale parameter used to define correlation between previous and proposed
+        sample in random-walk.  *Optional*. If float, the same scale is used for
+        all dimensions. If ndarray, a (possibly) different scale is used for
+        each dimension.
+
+    initial_point : ndarray
+        Initial parameters. *Optional*
+
+    dim : int
```
nabriis, line 34: I would suggest we remove the dim property. Instead user can define UserDefinedDistribution. We can discuss at next meeting.

cuqi/experimental/mcmc/_cwmh.py
```diff
@@ -0,0 +1,207 @@
...
+                 initial_point=None, **kwargs):
+        super().__init__(target, proposal=proposal, scale=scale,
+                         initial_point=initial_point, **kwargs)
+        # Set scale
+        if isinstance(scale, Number):
+            self.scale = np.ones(self.dim)*scale
+            self._scale_temp = np.ones(self.dim)*scale
+        elif isinstance(scale, np.ndarray):
+            self.scale = scale
+            self._scale_temp = scale.copy()
```
nabriis, line 84: We may want to include this in scale setter method for robustness?

cuqi/experimental/mcmc/_cwmh.py
```diff
@@ -0,0 +1,207 @@
...
+
+        # set initial acceptance rate
+        self._acc = [np.ones((self.dim))]
+
+    def validate_target(self):
+        pass # All targets are valid
+
+    @ProposalBasedSamplerNew.proposal.setter
```
nabriis, line 92: Consider adding a `#TODO` here related to issue 378. This overwriting of the proposal setter is a design we may reconsider.

cuqi/experimental/mcmc/_cwmh.py
```diff
@@ -0,0 +1,207 @@
...
+
+        elif not isinstance(value, cuqi.distribution.Distribution) and callable(
+            value):
+            self._proposal = value
+
+        else:
+            raise ValueError(fail_msg)
+
+    def step(self):
+        """ Perform one step of the sampler by transitioning the current point
```
nabriis, line 127: I believe we can push this generic docstring to the base class. Other classes will inherit the docstring automatically.

cuqi/experimental/mcmc/_cwmh.py
```diff
@@ -0,0 +1,207 @@
...
+            x_star = x_t.copy()
+
+        self.current_target = target_eval_t
+        self.current_point = x_t
+        # NEW: update return
+        # return x_t, target_eval_t, acc
+        return acc
+
+    def tune(self, skip_len, update_count):
+        """ Tune the parameters of the sampler. This method is called after each
```
nabriis, line 171: Same comment about putting the generic docstring to base class

tests/experimental/test_mcmc.py
```diff
@@ -103,6 +114,31 @@ def test_MALA_regression_warmup(target: cuqi.density.Density):
...
+    """Test the CWMH sampler regression."""
+    sampler_old = cuqi.sampler.CWMH(target, scale=np.ones(target.dim))
+    sampler_new = cuqi.experimental.mcmc.CWMHNew(target,
+                                                 scale=np.ones(target.dim))
+    Ns = 100 if target.dim < 50 else 20
+    assert_true_if_warmup_is_equivalent(sampler_old, sampler_new,
+                                        Ns=Ns,
+                                        strategy="MH_like",
+                                        old_idx=[0, -1],
+                                        new_idx=[1, None])
```
nabriis, line 141: Instead of old_idx and new_idx, which implicitly defines a range of values. I suggest for readabiltiy that we define an explicit range that we compare for the two samplers. E.g. one could compare [1..N] with [0..N-1] or similar. I think it would make the test much easier to understand.

cuqi/experimental/mcmc/_cwmh.py
```diff
@@ -0,0 +1,207 @@
...
+                pass
+                # x_t[j]       = x_t[j]
+                # target_eval_t = target_eval_t
+            x_star = x_t.copy()
+
+        self.current_target = target_eval_t
+        self.current_point = x_t
+        # NEW: update return
+        # return x_t, target_eval_t, acc
+        return acc
```
nabriis, line 168: In a way I like that you copied the sampler "step" explicitly. But since we are comparing with regression tests, I think it would be worthwhile to "clean up" some of the comments and improve the overall code quality of the step method in this new module.

Same comment holds for the warmup method.

nabriis (changes requested):
Hi @amal-ghamdi. Thanks for adding the CWMH sampler in the module. The suggested changes look good overall, I simply had some minor comments in relation to the testing framework. In addition, I also had a comment on cleaning up the step and tune methods even more compared to the Sampler module.

cuqi/experimental/mcmc/_cwmh.py
amal-ghamdi: good idea! done

cuqi/experimental/mcmc/_cwmh.py
amal-ghamdi: done

cuqi/experimental/mcmc/_cwmh.py
amal-ghamdi: yes, that is great. I removed the doc string in step, tune, set_state and get_state.

cuqi/experimental/mcmc/_cwmh.py
amal-ghamdi: Nice, I updated the code for better readability, style and documentation

tests/experimental/test_mcmc.py
amal-ghamdi: I agree, this makes it clearer. I actually thought about this, but it can make it tricky to set default sample size and default range since the range depends on the sample size. Also setting Ns and range separately, adds overhead on the developer to make them consistent, i.e. sat for the same N.

amal-ghamdi:
Thank you @nabriis for your helpful review! I addressed all your comments except the one about the dimension. We can talk about it today. I also added a comment about your suggested test change: from index to range.

cuqi/experimental/mcmc/_cwmh.py
nabriis: After our discussion, we agreed to remove the "dim" property yes.

tests/experimental/test_mcmc.py
nabriis: Right. I think I got a bit confused. Perhaps this approach of using the indices for each set of samples makes sense. It may be helpful to add docstring with the parameters of this test.

cuqi/experimental/mcmc/_cwmh.py
```diff
@@ -0,0 +1,222 @@
...
+        x_star = self.current_point.copy()
+
+        # Propose a sample x_all_components from the proposal distribution
+        # for all the components
+        target_eval_t = self.current_target
+        if isinstance(self.proposal,cuqi.distribution.Distribution):
+            x_all_components = self.proposal(
+                location= self.current_point, scale=self.scale).sample()
+        else:
+            x_all_components = self.proposal(self.current_point, self.scale)
```
nabriis, line 153: This may be out of scope for this PR, but I wonder if we should get rid of the callable proprosal also? And just use distributions. Could create issue for this.

cuqi/experimental/mcmc/_cwmh.py
nabriis: Thanks! It looks much better now.

nabriis (approved):
Thanks @amal-ghamdi. Looks good to me! I only had some small comments. I can approve it from here.

cuqi/experimental/mcmc/_cwmh.py
amal-ghamdi: good point, I added it to #373

tests/experimental/test_mcmc.py
amal-ghamdi: Great, I added documentation for these parameters

amal-ghamdi:
Thank you @nabriis for your second review! I updated the PR to reflect your suggestions.

cuqi/experimental/mcmc/_cwmh.py
```diff
@@ -0,0 +1,221 @@
...
+
+    initial_point : ndarray
+        Initial parameters. *Optional*
+
+    callback : callable, *Optional*
+        If set this function will be called after every sample.
+        The signature of the callback function is
+        `callback(sample, sample_index)`, where `sample` is the current sample
+        and `sample_index` is the index of the sample.
+        An example is shown in demos/demo31_callback.py.
```
jakobsj, line 39: Is the way callbacks are illustrated in demo31 still relevant for the new samplers? Maybe add an issue to update the demo once the samplers have replaced the old samplers.

cuqi/experimental/mcmc/_cwmh.py
```diff
@@ -0,0 +1,221 @@
...
+    def __init__(self, target: cuqi.density.Density, proposal=None, scale=1,
+                 initial_point=None, **kwargs):
+        super().__init__(target, proposal=proposal, scale=scale,
+                         initial_point=initial_point, **kwargs)
+
+        # set initial scale
+        self.scale = scale
+
+        # set initial acceptance rate
+        self._acc = [np.ones((self.dim))]
```
jakobsj, line 79: Is `dim` still available?  I understood from the PR comments it had been removed?

cuqi/experimental/mcmc/_cwmh.py
```diff
@@ -0,0 +1,221 @@
...
+        self._scale_temp = self._scale.copy()
+
+    def validate_target(self):
+        if not isinstance(self.target, cuqi.density.Density):
+            raise ValueError(
+                "Target should be an instance of "+\
+                f"{cuqi.density.Density.__class__.__name__}")
+
+    @ProposalBasedSamplerNew.proposal.setter
+    # TODO. Check if we can refactor this.
```
jakobsj, line 102: Do we have an issue to remember this? If not, add one?

jakobsj (approved):
Thanks @amal-ghamdi , from a quick look through I think it looks great and only have a couple of tiny questions added as individual comments, but approving now.

cuqi/experimental/mcmc/_cwmh.py
amal-ghamdi: good point, I recorded this in #378

cuqi/experimental/mcmc/_cwmh.py
amal-ghamdi: `dim` as parameter to the initializer is removed. But we still have property `dim` extracted from the target dim.

cuqi/experimental/mcmc/_cwmh.py
amal-ghamdi: Thank you for pointing this out, I recorded it in #373

amal-ghamdi:
@jakobsj , Many thanks for your review! I addressed your comments, feel free to resolve the conversations if you think the comments are addressed properly.

Issue #367, opened 2024-03-04
Enable Gibbs sampling in experimental mcmc module


amal-ghamdi:
New GibbsSampler based on SamplerNew
Allowing target=None

PR #382, opened 2024-03-11
Sampler rework LinearRTONew and RegularizedLinearRTONew

### LINKED ISSUE #368: Subclass the LinearRTO/regularizedLinearRTO samplers from the new Sampler class

chaozg:
fixed #368 

- [x] LinearRTO
  - [x] overwrite setter for target
  - [x] put precomputation-relevant code in _precompute
  - [x] put _precompute into target setter
- [x] RegularizedLinearRTO
  - [x] move stepsize-relevant code in _choose_stepsize
  - [x] check if prior has proximal in target setter 
 - [x] Unit tests with both sample and warmup
   - [x] plain/usual targets with LinearRTO
   - [x] MultipleLikelihoodPosterior targets with LinearRTO
   - [x] regularized targets with RegularizedLinearRTO
   - [x] regularized MultipleLikelihoodPosterior targets with RegularizedLinearRTO


cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,265 @@
...
+                    for likelihood in self.likelihoods:
+                        idx_end += len(likelihood.data)
+                        out1 += likelihood.model.adjoint(likelihood.distribution.sqrtprec.T@x[idx_start:idx_end])
+                        idx_start = idx_end
+                    out2 = L2.T @ x[idx_end:]
+                    out  = out1 + out2                
+                return out   
+            self.M = M  
+        else:
+            raise TypeError("All likelihoods need to be callable or none need to be callable.")
```
nabriis, line 122: As I know we will need to support initializing with `target=None`I suggest we add this behavior to a target.setter method.

You can look at the target.setter of the base class `SamplerNew` and re-define it in this class.

Essentially my thought is that whenever `target` is set, all these steps need to be carried out again **except setting current_point**.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,265 @@
...
+
+    @property
+    def likelihood(self):
+        return self.target.likelihood
+    
+    @property
+    def likelihoods(self):
+        if isinstance(self.target, cuqi.distribution.Posterior):
+            return [self.target.likelihood]
+        elif isinstance(self.target, cuqi.distribution.MultipleLikelihoodPosterior):
```
nabriis, line 136: I think we should add a regression test for these samplers with the MultipleLikelihoodPosterior.

Could be done by testing on the sum of two lists of posteriors. The `basic_targets` and `RTO_targets`

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,265 @@
...
+    @property
+    def data(self):
+        return self.target.data
+
+    def step(self):
+        y = self.b_tild + np.random.randn(len(self.b_tild))
+        sim = CGLS(self.M, y, self.current_point, self.maxit, self.tol, self.shift)            
+        self.current_point, _ = sim.solve()
+        acc = 1
+        return acc
```
nabriis, line 152: Step is nice and simple :)

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,265 @@
...
+                    idx_end = 0
+                    out1 = np.zeros(self.n)
+                    for likelihood in self.likelihoods:
+                        idx_end += len(likelihood.data)
+                        out1 += likelihood.model.adjoint(likelihood.distribution.sqrtprec.T@x[idx_start:idx_end])
+                        idx_start = idx_end
+                    out2 = L2.T @ x[idx_end:]
+                    out  = out1 + out2                
+                return out   
+            self.M = M
```
nabriis, line 120: While I know it was also part of the older LinearRTO samplers, I was thinking if we could improve the naming of some variables here? And also make most of them private.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,265 @@
...
+        sim = CGLS(self.M, y, self.current_point, self.maxit, self.tol, self.shift)            
+        self.current_point, _ = sim.solve()
+        acc = 1
+        return acc
+
+    def tune(self, skip_len, update_count):
+        pass
+
+    def _sample_adapt(self, N, Nb):
+        return self._sample(N,Nb)
```
nabriis, line 158: What is this doing here?

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,265 @@
...
+                        idx_start = idx_end
+                    out2 = L2.T @ x[idx_end:]
+                    out  = out1 + out2                
+                return out   
+            self.M = M  
+        else:
+            raise TypeError("All likelihoods need to be callable or none need to be callable.") 
+
+    @property
+    def prior(self):
```
nabriis, line 125: Let us make all of these variables private. Hopefully should be fairly easy in VS code with F2.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,265 @@
...
+            if not hasattr(self.likelihood.distribution, "sqrtprec"):
+                raise TypeError("Distribution in Likelihood must contain a sqrtprec attribute")
+            
+        elif isinstance(self.target, cuqi.distribution.MultipleLikelihoodPosterior): # Elif used for further alternatives, e.g., stacked posterior
+            for likelihood in self.likelihoods:
+                if not isinstance(likelihood.model, cuqi.model.LinearModel):
+                    raise TypeError("Model needs to be linear")
+
+                if not hasattr(likelihood.distribution, "sqrtprec"):
+                    raise TypeError("Distribution in Likelihood must contain a sqrtprec attribute")
```
nabriis, line 179: Feels to me like some of this behavior like `likelihood` and `likelihoods` could be simplified. Perhaps having `likelihoods` return even for 1 likelihood?

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,265 @@
...
+    Regularized Linear RTO (Randomize-Then-Optimize) sampler.
+
+    Samples posterior related to the inverse problem with Gaussian likelihood and implicit Gaussian prior, and where the forward model is Linear.
+
+    Parameters
+    ------------
+    target : `cuqi.distribution.Posterior`
+        See `cuqi.sampler.LinearRTO`
+
+    x0 : `np.ndarray`
```
nabriis, line 205: x0 not supported. Its initial_point

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,265 @@
...
+        If set this function will be called after every sample.
+        The signature of the callback function is `callback(sample, sample_index)`,
+        where `sample` is the current sample and `sample_index` is the index of the sample.
+        An example is shown in demos/demo31_callback.py.
+        
+    """
+    def __init__(self, target, initial_point=None, maxit=100, stepsize = "automatic", abstol=1e-10, adaptive = True, **kwargs):
+
+        if not callable(target.prior.proximal):
+            raise TypeError("Projector needs to be callable")
```
nabriis, line 228: Add to target setter after other suggested update.

tests/zexperimental/test_mcmc.py
```diff
@@ -103,6 +103,36 @@ def test_MALA_regression_warmup(target: cuqi.density.Density):
...
+    A, y_data, info = cuqi.testproblem.Deconvolution1D(dim=n, phantom='square').get_components()
+    x = cuqi.implicitprior.RegularizedGaussian(0.5*np.ones(n), 0.1, constraint = "nonnegativity")
+    y = cuqi.distribution.Gaussian(A@x, 0.001)
+
+    joint = cuqi.distribution.JointDistribution(x, y)
+    target = joint(y=y_data)
+
+    sampler_old = cuqi.sampler.RegularizedLinearRTO(target)
+    sampler_new = cuqi.experimental.mcmc.RegularizedLinearRTONew(target)
+    assert_true_if_sampling_is_equivalent(sampler_old, sampler_new)
```
nabriis, line 135: I suggest creating another list of `targets` related to LinearRTO and RegularizedLinearRTO and using pytest parametrize to test against them.

I also suggest adding the `warmup` test even if it seems redundant with  `tune` being pass. It will catch any issues in the future.

nabriis (changes requested):
Quick review. Loosk really good @chaozg. It is very satisfying to see RegularizedLinearRTO be subclassed from LinearRTO! 👍

cuqi/experimental/mcmc/_rto.py
nabriis: Lets postpone this to another issue #389  to not derail the PR too much!

cuqi/experimental/mcmc/_rto.py
nabriis: You can leave this out of the PR. I can manage it in #367

cuqi/experimental/mcmc/_rto.py
chaozg: Thanks, I'll only put the code in `self._precompute` for now

cuqi/experimental/mcmc/_rto.py
chaozg: Totally agree : )

cuqi/experimental/mcmc/_rto.py
chaozg: Sorry for being careless: it was copied for the old sampler class and is deleted now

cuqi/experimental/mcmc/_rto.py
chaozg: I'm make them private later in #389

cuqi/experimental/mcmc/_rto.py
chaozg: Sorry Nicolai, I don't get your point here: to my understanding, the code you highlighted is to check if the provided `Posterior`/`MultipleLikelihoodPosterior` is connected to a linear model. Do we expect any return here?

cuqi/experimental/mcmc/_rto.py
chaozg: Done

cuqi/experimental/mcmc/_rto.py
chaozg: I move this part to the setter of RegularizedLinearRTO's `target`

tests/zexperimental/test_mcmc.py
chaozg: Just added a function to create a list of "regularized" targets to test with RegularizedLinearRTO. Relevant code is like:
```python
def create_regularized_target(dim=16):
    """Create a regularized target."""
    A, y_data, info = cuqi.testproblem.Deconvolution1D(dim=dim, phantom='square').get_components()
    x = cuqi.implicitprior.RegularizedGaussian(0.5*np.ones(dim), 0.1, constraint = "nonnegativity")
    y = cuqi.distribution.Gaussian(A@x, 0.001)
    return cuqi.distribution.JointDistribution(x, y)(y=y_data)

regularized_targets = [
    create_regularized_target(dim=32),
    create_regularized_target(dim=64),
    create_regularized_target(dim=128)
]
```

cuqi/experimental/mcmc/_rto.py
chaozg: Done

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,273 @@
...
+
+        # Other parameters
+        self.stepsize = stepsize
+        self.abstol = abstol   
+        self.adaptive = adaptive
+        self.proximal = target.prior.proximal
+        self._stepsize = self._choose_stepsize()
+        self.maxit = maxit
+
+    @LinearRTONew.target.setter
```
nabriis, line 241: Same comment here.

tests/zexperimental/test_mcmc.py
```diff
@@ -156,6 +156,128 @@ def test_MALA_regression_warmup(target: cuqi.density.Density):
     sampler_new = cuqi.experimental.mcmc.MALANew(target, scale=1)
     assert_true_if_warmup_is_equivalent(sampler_old, sampler_new)
 
+# ============ LinearRTO ============
+
+@pytest.mark.parametrize("target", targets)
```
nabriis, line 161: I was thinking you could test on targets+multiple_likelihood_posterior_targets here. Then remove the redundant lines below

nabriis (changes requested):
Hi @chaozg. Looks very good! The PR is close to ready in my view now. Only some very minor comments and some suggestions for refactoring the tests.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,273 @@
...
+            # Prior TODO: allow multiple priors stacked
+            #if isinstance(P_mean, list) and isinstance(P_sqrtprec, list):
+            #    P = cuqi.distribution.JointGaussianSqrtPrec(P_mean, P_sqrtprec)
+            #else:
+            P = cuqi.distribution.Gaussian(P_mean, sqrtprec=P_sqrtprec)
+
+            # Construct posterior
+            value = cuqi.distribution.Posterior(L, P)
+        self._target = value
+        self.validate_target()
```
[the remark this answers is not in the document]
chaozg: Hi @nabriis , it turns out `super().target = value` alone doesn't work here, and I used
`super(LinearRTONew, type(self)).target.fset(self, value)` instead

tests/zexperimental/test_mcmc.py
chaozg: Done as suggested; I defined a `LinearRTO_targets` as:
```python
LinearRTO_targets = targets + [
    create_MultipleLikelihoodPosterior_target(dim=32),
    create_MultipleLikelihoodPosterior_target(dim=64),
    create_MultipleLikelihoodPosterior_target(dim=128)
]
```

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,273 @@
...
+        self.callback = callback
+
+        self._samples = [np.zeros(self.dim)] #TODO: Should we store zeros as first sample?
+        self.current_point = self.initial_point
+        self._acc = [1] # TODO. Check if we need this
+
+        # Other parameters
+        self.maxit = maxit
+        self.tol = tol        
+        self.shift = shift
```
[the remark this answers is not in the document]
nabriis: Ok. Seems good for now. I think we will rework the initial point handling to a more general method soon.

nabriis (approved):
Thanks @chaozg. LGTM now :)

jakobsj (changes requested):
Thank you @chaozg for this very nice reimplementation of LinearRTO samplers. It looks good, in particular how the regularized version is now subclassed from the standard version. Thanks also to @nabriis for the very thorough reviews. I don't have many additional comments, only added a few minor ones.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,273 @@
...
+        Tolerance of the inner CGLS solver. *Optional*.
+
+    callback : callable, *Optional*
+        If set this function will be called after every sample.
+        The signature of the callback function is `callback(sample, sample_index)`,
+        where `sample` is the current sample and `sample_index` is the index of the sample.
+        An example is shown in demos/demo31_callback.py.
+        
+    """
+    def __init__(self, target, maxit=10, tol=1e-6, shift=0, **kwargs):
```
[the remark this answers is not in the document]
chaozg: Done

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,273 @@
...
+                model = cuqi.model.LinearModel(model)
+
+            # Check model input
+            if not isinstance(model, cuqi.model.LinearModel):
+                raise TypeError("Model needs to be cuqi.model.LinearModel or matrix")
+
+            # Likelihood
+            L = cuqi.distribution.Gaussian(model, sqrtprec=L_sqrtprec).to_likelihood(data)
+
+            # Prior TODO: allow multiple priors stacked
```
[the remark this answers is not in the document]
chaozg: I just created [#400](https://github.com/CUQI-DTU/CUQIpy/issues/400) for this future feature

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,273 @@
...
+        Absolute tolerance of the inner FISTA solver. *Optional*.
+
+    callback : callable, *Optional*
+        If set this function will be called after every sample.
+        The signature of the callback function is `callback(sample, sample_index)`,
+        where `sample` is the current sample and `sample_index` is the index of the sample.
+        An example is shown in demos/demo31_callback.py.
+        
+    """
+    def __init__(self, target, maxit=100, stepsize = "automatic", abstol=1e-10, adaptive = True, **kwargs):
```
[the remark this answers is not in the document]
chaozg: - Added description on `adaptive` in docstring.
- As to `initial_point`, I think it is taken care of as part of `kwargs`, then we don't need to explicitly list it as an argument of the constructor? I'm confident here.

cuqi/experimental/mcmc/_sampler.py
```diff
@@ -83,6 +83,11 @@ def initial_point(self):
         """ Return the initial point of the sampler. This is always the first sample. """
         return self._samples[0]
     
+    @initial_point.setter
+    def initial_point(self, value):
+        """ Set the initial point of the sampler. """
+        self._samples[0] = value
```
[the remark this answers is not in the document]
chaozg: - The setter for `initial_point` in base class `SamplerNew` is missing, while we need to set `initial_point` in LinearRTO: that's why we added this setter in the base, as suggested by Nicolai.
- As to if the initial point should be stored as a sample, I think it's not decided yet. I created issue #401 to record this problem.

cuqi/experimental/mcmc/_rto.py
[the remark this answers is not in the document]
jakobsj: Right, I see it is handled through the super call by the lines of LinearRTONew:
```
        if 'initial_point' not in kwargs:
            self.initial_point = np.zeros(self.dim)
```
Two questions then. 

Can this default value not be specified as as part of the `__init__` line? It would seem neater that `initial_point` is seen explicitly in the list of inputs, such that the list of inputs matches exactly the inputs stated in the docstring. Is it because `self.dim` needs to be set before we can create the zero vector of the correct size, and setting `self.dim` is handled by the `super` call from `LinearRTO`?

The `__init__` inputs to `LinearRTONew` and `RegularizedLinearRTONew` are slightly different
```
def __init__(self, target, maxit=10, tol=1e-6, shift=0, **kwargs)
```
vs
```
    def __init__(self, target, maxit=100, stepsize = "automatic", abstol=1e-10, adaptive = True, **kwargs):
```
Is that because, in the first one, `tol` and `shift` are just used for the CGLS, while for the second `stepsize`, `abstol` and `adaptive` are just used for FISTA?

cuqi/experimental/mcmc/_rto.py
```diff
@@ -0,0 +1,279 @@
...
+        If True, FISTA is used as inner solver, otherwise ISTA is used. *Optional*.
+
+    callback : callable, *Optional*
+        If set this function will be called after every sample.
+        The signature of the callback function is `callback(sample, sample_index)`,
+        where `sample` is the current sample and `sample_index` is the index of the sample.
+        An example is shown in demos/demo31_callback.py.
+        
+    """
+    def __init__(self, target, maxit=100, stepsize = "automatic", abstol=1e-10, adaptive = True, **kwargs):
```
jakobsj, line 235:
```suggestion
    def __init__(self, target, maxit=100, stepsize="automatic", abstol=1e-10, adaptive=True, **kwargs):
```

jakobsj (changes requested):
Thanks for the update. Most resolved, only one small additional change suggestion on whitespace, and one unresolved conversation with a couple of questions.

cuqi/experimental/mcmc/_rto.py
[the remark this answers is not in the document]
chaozg: Done as suggested after group discussion

jakobsj (approved):

PR #385, opened 2024-03-11
Make progressbar2 dependency not explicit

### LINKED ISSUE #384: CUQIpy main now requires progressbar2
**Description**

CUQIpy main now requires progressbar due to its experimental module.

**Example to reproduce**

Do not have progressbar2 installed. Run this

```python
import cuqi
```

Gives error that module is not installed.

nabriis:
Closes #384


chaozg (approved):
LGTM

PR #387, opened 2024-03-13
Sampler rework: NUTS

amal-ghamdi:


cuqi/experimental/mcmc/_hmc.py
```diff
@@ -0,0 +1,401 @@
...
+
+    Parameters
+    ----------
+    target : `cuqi.distribution.Distribution`
+        The target distribution to sample. Must have logpdf and gradient method.
+        Custom logpdfs and gradients are supported by using a
+        :class:`cuqi.distribution.UserDefinedDistribution`.
+    
+    initial_point : ndarray
+        Initial parameters. *Optional*
```
jakobsj, line 25: Since optional, specify what happens if the user does not specify the initial_point.

cuqi/experimental/mcmc/_hmc.py
```diff
@@ -0,0 +1,401 @@
...
+
+    max_depth : int
+        Maximum depth of the tree.
```
jakobsj, line 28: State this should be integer greater than 0(?)

cuqi/experimental/mcmc/_hmc.py
```diff
@@ -0,0 +1,401 @@
...
+
+    adapt_step_size : Bool or float
+        Whether to adapt the step size.
+        If True, the step size is adapted automatically.
+        If False, the step size is fixed to the initially estimated value.
+        If set to a scalar, the step size will be given by user and not adapted.
```
jakobsj, line 34: Specify any constraints/range for the scalar. Nonnegative? Smaller than 1?

cuqi/experimental/mcmc/_hmc.py
```diff
@@ -0,0 +1,401 @@
...
+
+    opt_acc_rate : float
+        The optimal acceptance rate to reach if using adaptive step size.
+        Suggested values are 0.6 (default) or 0.8 (as in stan).
```
jakobsj, line 38: State the range for this parameter. Greater than (or equal?) zero? Smaller than (or equal?) to 1.0?

cuqi/experimental/mcmc/_hmc.py
```diff
@@ -0,0 +1,401 @@
...
+
+        # Import cuqi
+        import cuqi
+
+        # Define a target distribution
+        tp = cuqi.testproblem.WangCubic()
+        target = tp.posterior
+
+        # Set up sampler
+        sampler = cuqi.experimental.mcmc.NUTSNew(target)
```
jakobsj, line 60: Nice example, also showing how the diagnostics can be accessed after sampling.

cuqi/experimental/mcmc/_hmc.py
```diff
@@ -0,0 +1,401 @@
...
+        eta1 = 1/(k + t_0)
+        self._H_bar = (1-eta1)*self._H_bar +\
+            eta1*(self.opt_acc_rate - (self._alpha/self._n_alpha))
+        self._epsilon = np.exp(self._mu - (np.sqrt(k)/gamma)*self._H_bar)
+        eta = k**(-kappa)
+        self._epsilon_bar =\
+            np.exp(eta*np.log(self._epsilon) +(1-eta)*np.log(self._epsilon_bar))
+
+    def get_state(self):
+        pass
```
jakobsj, line 223: Does this need updating to set and get specific keys for the state of NUTS?

cuqi/experimental/mcmc/_sampler.py
```diff
@@ -72,6 +72,14 @@ def validate_target(self):
         """ Validate the target is compatible with the sampler. Called when the target is set. Should raise an error if the target is not compatible. """
         pass
 
+    def _pre_sample(self):
```
jakobsj, line 75: On whether to make these abstract, what would be the pros and cons? If non-abstract, as now, it seems ok to me that only if the user (coder of a new sampler) specifically wants something to happen there, they will overwrite these methods. Otherwise, the current "pass" methods will simply be called.

jakobsj (commented):
THanks @amal-ghamdi I added my general comments and questions. For more technical review my suggestion would be to request from both @chaozg and @nabriis.

cuqi/experimental/mcmc/_hmc.py
```diff
@@ -0,0 +1,401 @@
...
+        # number of tree nodes created each NUTS iteration
+        self._num_tree_node = 0
+        # Create lists to store NUTS run diagnostics
+        self._create_run_diagnostic_attributes()
+
+    #=========================================================================
+    #================== Implement methods required by SamplerNew =============
+    #=========================================================================
+    def validate_target(self):
+        pass #TODO: target needs to have logpdf and gradient methods
```
nabriis, line 121: I would suggest adding a try catch here, simply attemping to evaluating _nuts_target and if the output is valid, accept the target.

cuqi/experimental/mcmc/_hmc.py
nabriis: If possible I would suggest adding the keys that NUTS uses for state to the class and removing the get_state and set_state methods.

It also gives  a clearer overview of what is the actual parameters that change during sampling.

cuqi/experimental/mcmc/_sampler.py
nabriis: I am fine with the _pre_sample and _pre_warmup being was they are here (that is not abstract). I can already tell we will use some general implementation, e.g. for "initial_point" in the base class very soon. Then I would suggest calling `super()._pre_warmup()` and `super()._pre_sample()` in the subclass though.

tests/zexperimental/test_mcmc.py
```diff
@@ -265,6 +272,35 @@ def test_CWMH_regression_warmup(target: cuqi.density.Density):
                                         old_idx=[0, -1],
                                         new_idx=[1, None])
 
+# ============= HMC (NUTS) ==============
```
nabriis, line 275: After merging of https://github.com/CUQI-DTU/CUQIpy/pull/390 there may be some other tests to add the NUTS sampler in. Consider rebasing and adding those.

nabriis (changes requested):
Thanks @amal-ghamdi. This is looking really nice. I like many of the decisions taken in this PR. I added some small comments on the raised questions and a few minor things.

amal-ghamdi:
@jakobsj and @nabriis, Many thanks for your very helpful review and the good points you raised! I'll work on these today.

cuqi/experimental/mcmc/_hmc.py
amal-ghamdi: I added the required description. I also added a setter and getter to make sure the input value is of the correct format.

cuqi/experimental/mcmc/_hmc.py
amal-ghamdi: Thanks for the feedback. I updated one comment in the example about the value of _epsilon_bar

cuqi/experimental/mcmc/_hmc.py
amal-ghamdi: Thank you both for pointing this out! I missed removing the `get_state` and `set_state`. Now it is removed. 

I appended keys to the STATE and HISTORY data structures. 

My plan for the state to be thoroughly tested (making sure no state variable is missing) when checkpointing is addressed #404

cuqi/experimental/mcmc/_sampler.py
amal-ghamdi: Probably it is fine just to make it not abstract. The only advantage of making it abstract in my opinion is to alert the developer to take care of these two methods. Thanks for the input here.

cuqi/experimental/mcmc/_hmc.py
amal-ghamdi: Thank you for the suggestion, I added the try-catch phrase.

tests/zexperimental/test_mcmc.py
amal-ghamdi: The main was merged into this branch. I added NUTS to the skip tests for checkpointing. My plan is to address checkpointing in #404

amal-ghamdi:
@jakobsj and @nabriis, Many thanks for all the comments! I addressed your comments and I believe it really helped improve the robustness of the sampler.

Additionally, I changed a bit the meaning of `adapt_step_size`. See the comment above replying to @jakobsj . Now also it is named `step_size`.

I also made some updates to `_pre_warmup` and `_pre_sample` to ensure that if the user starts `warmup` after another `warmup`, for example, the values of the state variables are up to date and not reset. Related to this is a test I plan to do as part of the issue #404 (copied here for reference)
#### The test
```
warmup(Ns)
warmup(Ns)
```
is equivalent to 
```
warmup(2*Ns)
```

If you prefer I address this test as part of this PR, I can do that. I think we have not tested this for all samplers yet, we possibly can make it a separate issue.

cuqi/experimental/mcmc/_hmc.py
jakobsj: Sounds good like a good improvement.

jakobsj (approved):
I'm happy with the changes regarding my comments.

cuqi/experimental/mcmc/_hmc.py
nabriis: Great thanks!

cuqi/experimental/mcmc/_sampler.py
nabriis: LGTM. Jakob  can resolve if he agrees!

nabriis (approved):
THanks @amal-ghamdi. The updates LGTM. Thanks for the fast response time :)

Issue #378, opened 2024-03-09
Sampler rework refactoring


nabriis:
This issue tracks comments primarily set in

amal-ghamdi:
@nabriis, I just noticed that status of this issue is `Done`. Should this be in `backlog`? Thanks!

nabriis:
I did it now. Odd. I created it via the project plan

PR #390, opened 2024-03-17
Implement get_state, set_state, get_history, set_history in base class.

### LINKED ISSUE #379: Refactor and fix "get_state" and "set_state" in MCMC module

### LINKED ISSUE #402: Iterate on get_state and set_state handling in experimental mcmc module
**Issue**
Current implementation of get_state and set_state is handled in each subclass of `SamplerNew` and not tested. This issue is to address this by implementing get_state and set_state in the base class and add unit tests. It is another problem to re-consider/refactor this design. For that see #379.

**Proposed solution**
1) Implement a datastructure that can contain keys to variables that currently constitute the state of each of the subclassing samplers. The subclassing samplers should be able to modify this datastructure in a simple way and clear way when/if they add new variables to be stored as state.

2) Implement the functionality of `get_state` and `set_state` in the base class and delete it from subclasses.

3) Add unit tests on get_state, set_state and checkpointing.

nabriis:
As described in #402 this PR seeks to implement get_state, set_state in the base class ´SamplerNew`. 

This is achieved by adding a class variable `_STATE_KEYS` that contain a set of keys representing the state variables the sampler will work with when calling the **already** implemented `get_state` and `set_state` methods. This reduces the amount of repeated code in each subclass.

Subclassing samplers must now _optionally_ update `_STATE_KEYS` if they expand on the set of variables the represent the state. For example the `ProposalBasedSampler` sampler can expand this by including the `scale` key, as this sampler maintains a `scale` attribute that is updated when running warmup sampling.

A testing framework is added get_state, set_state and checkpointing.

Additionally the `_HISTORY_KEYS` class variable is also added following the same design, but containing variable that representing the sampling history like the samples and acceptance criteria. This specific change introduces public methods get_history and set_history working similarly to how get_state and set_state already works before this PR.

Closes #402

The following related issue #379 is about refactoring the general behavior of get_state and set_state. This current implemented changes introduce 2 class variables that can relatively easily be replaced with another design once we agree on something else for #379.

### DIFF
--- a/cuqi/experimental/mcmc/_cwmh.py
+++ b/cuqi/experimental/mcmc/_cwmh.py
@@ -144,7 +144,7 @@ class CWMHNew(ProposalBasedSamplerNew):
 
         # Propose a sample x_all_components from the proposal distribution
         # for all the components
-        target_eval_t = self.current_target
+        target_eval_t = self.current_target_logd
         if isinstance(self.proposal,cuqi.distribution.Distribution):
             x_all_components = self.proposal(
                 location= self.current_point, scale=self.scale).sample()
@@ -175,7 +175,7 @@ class CWMHNew(ProposalBasedSamplerNew):
 
             x_star = x_t.copy()
 
-        self.current_target = target_eval_t
+        self.current_target_logd = target_eval_t
         self.current_point = x_t
 
         return acc
@@ -207,7 +207,7 @@ class CWMHNew(ProposalBasedSamplerNew):
 
         return {'sampler_type': 'CWMH',
                 'current_point': current_point,
-                'current_target': self.current_target,
+                'current_target': self.current_target_logd,
                 'scale': self.scale}
 
     def set_state(self, state):
@@ -217,5 +217,5 @@ class CWMHNew(ProposalBasedSamplerNew):
                                       geometry=self.target.geometry)
 
         self.current_point = current_point
-        self.current_target = state['current_target']
+        self.current_target_logd = state['current_target']
         self.scale = state['scale']
--- a/cuqi/experimental/mcmc/_cwmh.py
+++ b/cuqi/experimental/mcmc/_cwmh.py
@@ -199,23 +199,3 @@ class CWMHNew(ProposalBasedSamplerNew):
         # Update the scale parameter
         self.scale = np.minimum(scale_temp, np.ones(self.dim))
         self._scale_temp = scale_temp
-
-    def get_state(self):
-        current_point = self.current_point
-        if isinstance(current_point, CUQIarray):
-            current_point = current_point.to_numpy()
-
-        return {'sampler_type': 'CWMH',
-                'current_point': current_point,
-                'current_target': self.current_target_logd,
-                'scale': self.scale}
-
-    def set_state(self, state):
-        current_point = state['current_point']
-        if not isinstance(current_point, CUQIarray):
-            current_point = CUQIarray(current_point,
-                                      geometry=self.target.geometry)
-
-        self.current_point = current_point
-        self.current_target_logd = state['current_target']
-        self.scale = state['scale']
--- a/cuqi/experimental/mcmc/_pcn.py
+++ b/cuqi/experimental/mcmc/_pcn.py
@@ -5,13 +5,15 @@ from cuqi.array import CUQIarray
 
 class pCNNew(SamplerNew):  # Refactor to Proposal-based sampler?
 
+    _STATE_KEYS = SamplerNew._STATE_KEYS + ['scale', 'current_loglike_logd']
+
     def __init__(self, target, scale=1.0, **kwargs):
 
         super().__init__(target, **kwargs)
 
         self.scale = scale
         self.current_point = self.initial_point
-        self.current_loglike_eval = self._loglikelihood(self.current_point)
+        self.current_loglike_logd = self._loglikelihood(self.current_point)
 
         self._acc = [1] # TODO. Check if we need this
 
@@ -33,7 +35,7 @@ class pCNNew(SamplerNew):  # Refactor to Proposal-based sampler?
         loglike_eval_star =  self._loglikelihood(x_star) 
 
         # ratio and acceptance probability
-        ratio = loglike_eval_star - self.current_loglike_eval  # proposal is symmetric
+        ratio = loglike_eval_star - self.current_loglike_logd  # proposal is symmetric
         alpha = min(0, ratio)
 
         # accept/reject
@@ -41,7 +43,7 @@ class pCNNew(SamplerNew):  # Refactor to Proposal-based sampler?
         u_theta = np.log(np.random.rand())
         if (u_theta <= alpha):
             self.current_point = x_star
-            self.current_loglike_eval = loglike_eval_star
+            self.current_loglike_logd = loglike_eval_star
             acc = 1
         
         return acc
@@ -87,15 +89,3 @@ class pCNNew(SamplerNew):  # Refactor to Proposal-based sampler?
 
     def tune(self, skip_len, update_count):
         pass
-
-    def get_state(self):
-        return {'sampler_type': 'PCN', 'current_point': self.current_point.to_numpy(), \
-                'current_loglike_eval': self.current_loglike_eval.to_numpy(), \
-                'scale': self.scale}
-
-    def set_state(self, state):
-        temp = CUQIarray(state['current_point'] , geometry=self.target.geometry)
-        self.current_point = temp
-        temp = CUQIarray(state['current_loglike_eval'] , geometry=self.target.geometry)
-        self.current_loglike_eval = temp
-        self.scale = state['scale']
--- a/cuqi/experimental/mcmc/_pcn.py
+++ b/cuqi/experimental/mcmc/_pcn.py
@@ -5,7 +5,7 @@ from cuqi.array import CUQIarray
 
 class pCNNew(SamplerNew):  # Refactor to Proposal-based sampler?
 
-    _STATE_KEYS = SamplerNew._STATE_KEYS + ['scale', 'current_likelihood_logd']
+    _STATE_KEYS = SamplerNew._STATE_KEYS.union({'scale', 'current_likelihood_logd'})
 
     def __init__(self, target, scale=1.0, **kwargs):
 
--- a/cuqi/experimental/mcmc/_pcn.py
+++ b/cuqi/experimental/mcmc/_pcn.py
@@ -5,7 +5,7 @@ from cuqi.array import CUQIarray
 
 class pCNNew(SamplerNew):  # Refactor to Proposal-based sampler?
 
-    _STATE_KEYS = SamplerNew._STATE_KEYS + ['scale', 'current_loglike_logd']
+    _STATE_KEYS = SamplerNew._STATE_KEYS + ['scale', 'current_likelihood_logd']
 
     def __init__(self, target, scale=1.0, **kwargs):
 
@@ -13,7 +13,7 @@ class pCNNew(SamplerNew):  # Refactor to Proposal-based sampler?
 
         self.scale = scale
         self.current_point = self.initial_point
-        self.current_loglike_logd = self._loglikelihood(self.current_point)
+        self.current_likelihood_logd = self._loglikelihood(self.current_point)
 
         self._acc = [1] # TODO. Check if we need this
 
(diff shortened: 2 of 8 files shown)


cuqi/experimental/mcmc/_cwmh.py
```diff
@@ -199,23 +199,3 @@ def tune(self, skip_len, update_count):
         # Update the scale parameter
         self.scale = np.minimum(scale_temp, np.ones(self.dim))
         self._scale_temp = scale_temp
-
-    def get_state(self):
```
amal-ghamdi, line 203: really great how much redundancy were removed (the definitions of get_state and set_state that were everywhere)!

cuqi/experimental/mcmc/_pcn.py
```diff
@@ -5,13 +5,15 @@
...
+    _STATE_KEYS = SamplerNew._STATE_KEYS + ['scale', 'current_loglike_logd']
+
     def __init__(self, target, scale=1.0, **kwargs):
 
         super().__init__(target, **kwargs)
 
         self.scale = scale
         self.current_point = self.initial_point
-        self.current_loglike_eval = self._loglikelihood(self.current_point)
+        self.current_loglike_logd = self._loglikelihood(self.current_point)
```
amal-ghamdi, line 16: suggest changing this variable to `current_likelihood_logd` (changing all occurrences of the variable) to be consistent with `current_target_logd`

cuqi/experimental/mcmc/_sampler.py
```diff
@@ -13,6 +13,27 @@ def progressbar(iterable, **kwargs):
         warnings.warn("Module mcmc: Progressbar not found. Install progressbar2 to get sampling progress.")
         return iterable
 
+class _UniqueList(list):
```
amal-ghamdi, line 16: Is there a particular need for this data structure, instead of using Python set. for example:
```
set1 = {'key1'} # set elements are 'key1'
set1.add('key1') #  set elements are 'key1'
set1.add('key2') #  set elements are 'key1', 'key2'
set1.update({'key2', 'key3'})  #  set elements are 'key1', 'key2', 'key3'
```

cuqi/experimental/mcmc/_sampler.py
```diff
@@ -45,8 +72,10 @@ def __init__(self, target: cuqi.density.Density, initial_point=None, callback=No
         # Choose initial point if not given
         if initial_point is None:
             initial_point = np.ones(self.dim)
+
+        self.initial_point = initial_point
         
-        self._samples = [initial_point]
+        self._samples = [initial_point] # Remove?
```
amal-ghamdi, line 78: Do we need to capture this in one of the issues we have or create a new issue (Consider whether to include initial_point in the initial state of self._samples?)

cuqi/experimental/mcmc/_sampler.py
```diff
@@ -104,6 +117,25 @@ def target(self, value):
...
+
+    @property
+    def current_target_logd(self):
+        """ The current target value of the sampler. """
+        return self._current_target_logd
+    
+    @current_target_logd.setter
+    def current_target_logd(self, value):
+        """ Set the current target value of the sampler. """
+        self._current_target_logd = value
```
amal-ghamdi, line 139: This comment probably not necessarily should be addressed in this PR. I think we need a way to make sure developers who create subclass of NewSampler do use these two properties. Here are two suggestion:
1- to enforce these as abstractmethods. To prevent a case when developer make a subclass of NewSampler and not consider using these properties at all. That is, it is part of the interface the developer needs to implement. This is I think is a tradeoff. It will lead to some redundancy, but it will make it clear that the developer need to take care of 'step' 'tune' and the two states: 'current_point', ' current_target_logd'.
2- as we did before, ask user to implement `_tune` and `_step` instead of `tune` and `step`. Then we have parent method `tune` and `step` to call `_tune` / `_step` expecting `current_point` and `target_logd` to be returned. and store them in state variable. So it is not the responsibility of  `_tune` / `_step` to store this state variable.

cuqi/experimental/mcmc/_sampler.py
```diff
@@ -204,6 +236,54 @@ def warmup(self, Nb, tune_freq=0.1) -> 'SamplerNew':
             self._call_callback(self.current_point, len(self._samples)-1)
 
         return self
+    
+    def get_state(self) -> dict:
+        """ Return the state of the sampler. """   
+        state = {
+            'metadata': {
```
amal-ghamdi, line 243: it is good that you separated the metadata from what is required for state.

cuqi/experimental/mcmc/_sampler.py
```diff
@@ -204,6 +236,54 @@ def warmup(self, Nb, tune_freq=0.1) -> 'SamplerNew':
...
+            },
+            'state': {
+                key: getattr(self, key) for key in self._STATE_KEYS
+            }
+        }
+        return state
+
+    def set_state(self, state: dict):
+        """ Set the state of the sampler. """
+        if state['metadata']['sampler_type'] != self.__class__.__name__:
```
amal-ghamdi, line 254: nice :)

tests/experimental/test_mcmc.py
```diff
@@ -184,15 +185,54 @@ def test_CWMH_regression_warmup(target: cuqi.density.Density):
...
+    """Ensure that all samplers from cuqi.experimental.mcmc, except those skipped, are tested for checkpointing."""
+
+    # List of all samplers from cuqi.experimental.mcmc that should be tested for checkpointing
+    samplers = [
+        cls
+        for _, cls in inspect.getmembers(cuqi.experimental.mcmc, inspect.isclass)
+        if cls not in skip_checkpoint  # use cls here, not name
+    ]
+
+    # Convert instances to their classes
```
amal-ghamdi, line 214:
```suggestion
    # Convert instances in checkpoint_targets to their classes
```

tests/experimental/test_mcmc.py
```diff
@@ -202,16 +242,54 @@ def test_checkpointing(sampler: cuqi.experimental.mcmc.SamplerNew):
...
     # Do some more samples from pre-defined rng state
     np.random.seed(0)
-    samples2 = sampler_fresh.sample(100).get_samples().samples[...,1:] # TODO. This needs to be fixed..
+    samples2 = sampler_fresh.warmup(50).sample(50).get_samples().samples[...,1:] # TODO. This needs to be fixed.. We should likely not store initial point in _samples
 
     # Check that the samples are the same
     assert np.allclose(samples1, samples2), f"Samples1: {samples1}\nSamples2: {samples2}"
 
+
+@pytest.mark.parametrize("sampler", checkpoint_targets)
```
amal-ghamdi, line 259: we can probably make these 2 tests for state keys and history keys general to all samplers not only those that are tested for checkpointing.

amal-ghamdi (changes requested):
Thank you @nabriis for this great refactor! Looks much more robust than what we had. I added few comments.

jakobsj (commented):
Hi @nabriis, thank you for this, it is looking promising.  Like @amal-ghamdi I am wondering about the choice of data structure, whether a standard type like python set could be used. 

More generally, I was trying to work out how users should be able to interact with the state of samplers.  I think they should be able to easily view the current state of a sampler.  Load a state from disk, view it, perhaps modify, then set up a sampler with that state.  Perhaps even create a state from scratch (with easily seeing the expected entry types) and create a sampler.  I couldn't really see whether this was possible.  Is there a relevant demo outlining intended usage possibilities, to look at or can one be created?

It may be helpful to write out design/requirements to help see what the best data structure would be.

cuqi/experimental/mcmc/_pcn.py
nabriis: Done

cuqi/experimental/mcmc/_sampler.py
nabriis: done

cuqi/experimental/mcmc/_sampler.py
nabriis: Done

tests/experimental/test_mcmc.py
nabriis: Done

nabriis:
> Hi @nabriis, thank you for this, it is looking promising. Like @amal-ghamdi I am wondering about the choice of data structure, whether a standard type like python set could be used.
> 
> More generally, I was trying to work out how users should be able to interact with the state of samplers. I think they should be able to easily view the current state of a sampler. Load a state from disk, view it, perhaps modify, then set up a sampler with that state. Perhaps even create a state from scratch (with easily seeing the expected entry types) and create a sampler. I couldn't really see whether this was possible. Is there a relevant demo outlining intended usage possibilities, to look at or can one be created?
> 
> It may be helpful to write out design/requirements to help see what the best data structure would be.

I took your comment into account. The underlying datastructure to store the state internally in the sampler was updated to a set. The user is NOT exposed to this.

What the user is exposed to the what they can get from "get_state" and "set_state". These are dictionaries, and there is now more documentation on them.

The dictionary datastructure is useful because it is easy to extend what it can store (what keys it contains).

cuqi/experimental/mcmc/_sampler.py
nabriis: I agree there is something to consider here. I went with the following option:

Added a unit test to check if subclassing samplers utilize all state variables. See test_state_is_fully_updated_after_warmup_step.

We could still create an issue if you want to consider future, e.g. if users subclass within their ipynb or something.

We could add a "validate_sampler" method that can be run to check it works as expected.

cuqi/experimental/mcmc/_sampler.py
```diff
@@ -238,7 +206,28 @@ def warmup(self, Nb, tune_freq=0.1) -> 'SamplerNew':
...
+
+        For example, the state of a "MH" sampler could be:
+
+        state = {
+            'metadata': {
+                'sampler_type': 'MH'
+            },
+            'state': {
+                'current_point': np.array([...]),
+                'current_target_logd': -123.45,
```
amal-ghamdi, line 225: one thought here related to storing logd for all sampler. In some sense, logd is derived from `current_point`. If we want the sate to be minimal, maybe we restrict it to current point only without storing in logd values

tests/experimental/test_mcmc.py
```diff
@@ -293,3 +302,56 @@ def test_history_keys(sampler: cuqi.experimental.mcmc.SamplerNew):
 
     # Check if the actual keys match the expected keys
     assert actual_keys == expected_keys, f"History keys mismatch. Expected: {expected_keys}, Actual: {actual_keys}"
+
+# Dictionary to store keys that are not expected to be updated after warmup.
+# Likely due to not implemented feature in the sampler.
+state_exception_keys = {
```
amal-ghamdi, line 308: (this is not directly relevant to this PR) it seems all these classes allow warmup but tuning is not implemented yet (thus scale is not updated). I suggest we add an item in issue #378 to say :

- [ ] For all the sampler classes in which tune is implemented as 'pass', do we need warning or error? to let the user know tuning is not happening?

cuqi/experimental/mcmc/_sampler.py
amal-ghamdi: really nice added tests!

I just created this issue #392 for a general user-defined sampler (I do not think we have one). If you think it makes sense, you can add the details you suggested in this issue.

cuqi/experimental/mcmc/_sampler.py
nabriis: You are right. I moved the logd out of the base sampler class. It now currently recides in ProposalBasedSampler. It can be refactored further in the future.

amal-ghamdi (approved):
Thank you Nicolai, Looks great!

jakobsj:
Thanks! Please, for reference, add explanation as comment here on the PR on: 
- what you have done now in response to the latest comments and our discussions and decision around it, 
- what the latest commits seek to do, and 
- create an issue with explanation and the comments raised that have not been addressed by this PR and refer to it from here.

nabriis:
> Thanks! Please, for reference, add explanation as comment here on the PR on:
> 
> * what you have done now in response to the latest comments and our discussions and decision around it,
> * what the latest commits seek to do, and
> * create an issue with explanation and the comments raised that have not been addressed by this PR and refer to it from here.

Hello @jakobsj.

Thanks for requesting a clarification for future reference. I updated the issue #402 with a description of the sought change and updated this PR description with more details on the aim of this PR. Additionally, I included your comments into #379.

For reference this PR seeks to iterate on the current design by moving the get_state and set_state methods to the base class, not refactor the design in its entirety. Let us discuss that refactor in #379.

jakobsj (approved):
Many thanks for adding more explanation incl. scope, context and the capture of the review comments. This PR nicely addresses the issue of moving get/set state to the base class and managing state and history in a better way and provides a nice step forward toward at a later point addressing #379.

Issue #16, opened 2022-08-19
Add _logd (proportional) methods to distributions and make logpdf (exact) optional


nabriis:
Currently we require distributions to add a logpdf method. In actuality we often only need to know the logpdf up to a proportionality constant. For this reason we should instead require distributions to implement a log density (logd) method which returns the log probability density function up to a proportionality constant, and only optionally implement a logpdf method.

PR #420, opened 2024-05-06
Let Gaussian accept LinearOperator represented sqrtprec

### LINKED ISSUE #419: Let Gaussian accept LinearOperator as sqrtprec

chaozg:
fixed #419 

Issue: it might open some opportunities if we allow a Gaussian distribution to be initialized with `scipy.sparse.linalg.LinearOperator`. 

What I did:
- update `get_sqrtprec_from_sqrtcov` in `cuqi/distribution/_gaussian.py` to accept a`LinearOperator` object as `sqrtprec`;
- add `check_if_conditional_from_attr` in `cuqi/utilities/_utilities.py`, following @nabriis 's comment below;
- `check_if_conditional_from_attr` is used in `get_indirect_variables` so that a `Gaussian` distribution initialized with `LinearOperator` is not detected as a conditional distribution (just a reminder: a LinearOperator is callable);
- add unit test `test_Gaussian_from_linear_operator_sqrtprec` in `tests/test_distribution.py`

Some notes (please correct me if I'm wrong here):
- For MCMC methods in general, we don't need `logdet` as it will be cancelled when evaluating the density function of the proposal and the current state;
- In CUQIpy, we need `logdet` defined to compute `logpdf`: I think this is somehow unnecessary and the unnormalized version `_logupdf` should be enough in place of `logpdf`. But I have to admit that I'm _not confident_ if this point still holds when it comes to Gibbs sampling. (This issue turned out to be an echo from the old days #16, as Nicolai pointed in the comments below)
- In order to make a Gaussian distribution initialized with `LinearOperator` work with the current samplers in CUQIpy except ROTs, we need to set `logdet` to some value. It doesn't matter to what value it is set as it will be cancelled.
- For RTOs, we don't evaluate `logpdf`, so we don't need to set `logdet`.

### DIFF
--- a/cuqi/distribution/_gaussian.py
+++ b/cuqi/distribution/_gaussian.py
@@ -9,7 +9,7 @@ import scipy.linalg as splinalg
 
 from cuqi import config
 from cuqi.geometry import _get_identity_geometries
-from cuqi.utilities import force_ndarray, sparse_cholesky
+from cuqi.utilities import force_ndarray, sparse_cholesky, check_if_conditional_from_attr
 from cuqi.distribution import Distribution
 
 # We potentially allow the use of sksparse.cholmod for sparse Cholesky
@@ -191,7 +191,7 @@ class Gaussian(Distribution):
         value = force_ndarray(value)
         self._sqrtcov = value
         self._cov = None # Reset covariance (in case it was computed before)      
-        if (value is not None) and (not callable(value)): #TODO: problem here
+        if (value is not None) and (not callable(value)):
             if self.dim > config.MIN_DIM_SPARSE:
                 sparse_flag = True # do sparse computations
             else:
@@ -214,7 +214,7 @@ class Gaussian(Distribution):
         value = force_ndarray(value)
         self._sqrtprec = value
         self._cov = None # Reset covariance (in case it was computed before)
-        if (value is not None) and (not callable(value)):  
+        if not check_if_conditional_from_attr(value):
             if self.dim > config.MIN_DIM_SPARSE:
                 sparse_flag = True # do sparse computations
             else:
@@ -631,7 +631,7 @@ def get_sqrtprec_from_sqrtprec(dim, sqrtprec, sparse_flag):
     dim : int
         Dimension of the sqrtprec matrix.
     
-    sqrtprec : 1-d or 2-d ndarray or sparse matrix
+    sqrtprec : 1-d or 2-d ndarray or sparse matrix or scipy.sparse.linalg.LinearOperator
         Square root of precision matrix. If 1-dimensional, then assumed to be a diagonal matrix.
 
     sparse_flag: bool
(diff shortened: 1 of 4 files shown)


cuqi/distribution/_gaussian.py
```diff
@@ -673,6 +673,13 @@ def get_sqrtprec_from_sqrtprec(dim, sqrtprec, sparse_flag):
         logdet = np.sum(-np.log(precision))
         rank = dim
 
+    # sqrtprec is LinearOperator (only for RTOs)
+    elif isinstance(sqrtprec, spa.linalg.LinearOperator):
```
nabriis, line 677: I think this is a very reasonable addition. I would suggest adding to the existing unit tests cases for LinearOperator that check the logpdf values.

cuqi/distribution/_gaussian.py
```diff
@@ -673,6 +673,13 @@ def get_sqrtprec_from_sqrtprec(dim, sqrtprec, sparse_flag):
...
+        sqrtprec = sqrtprec
+        if hasattr(sqrtprec, 'logdet'):
```
nabriis, line 679: If logdet is not defined, what happens?

cuqi/utilities/_utilities.py
```diff
@@ -64,7 +64,7 @@ def get_indirect_variables(dist):
     attributes = []
     for attribute in dist.get_mutable_variables():
         value = getattr(dist, attribute)
-        if callable(value):
+        if callable(value) and not isinstance(value, spslinalg.LinearOperator):
```
nabriis, line 67: First "callable" object that we want to interpret as a fix variable. I suggest we add a method to check if "value" is one of the allowed cases that are callable but still act as a well-defined ndarray

nabriis (changes requested):
Hi @chaozg. I think it is a good idea to support LinearOperator. I had some sugggestions to add unit tests and also make the check in utilities more explicit by defining a method.

cuqi/distribution/_gaussian.py
chaozg: Done

cuqi/utilities/_utilities.py
chaozg: Added a method `check_if_conditional_from_attr` in `cuqi/utilities/_utilities.py` for your suggested job

cuqi/distribution/_gaussian.py
chaozg: It's really the user's choice/responsibility if they give `logdet` or not. If they give `logdet`, we will record it and then `logpdf` can be computed; if not, `logpdf` can't be computed. 

For large matrices, it's not easy to compute `logdet`, but luckily we don't need it for samplers like `LinearRTO`

chaozg:
Thanks @nabriis for your review and suggestions. I've updated the code to address your comments and am now requesting your further review.

cuqi/distribution/_gaussian.py
nabriis: This is OK for now.  I will create an issue related to the unormalized logpdf

cuqi/utilities/_utilities.py
nabriis: LGTM

nabriis (approved):
Thanks @chaozg. LGTM

chaozg:
Thanks @nabriis , I'm now requesting @jakobsj 's review.

cuqi/distribution/_gaussian.py
nabriis: There is an old issue related to this. See #16

chaozg: > There is an old issue related to this. See #16

Hahaha, it turned out to be an echo from two years ago : )

jakobsj (approved):
Nice addition and the workaround for logdet seems fine. I added a small question, you can merge after addressing this.

cuqi/distribution/_gaussian.py
```diff
@@ -666,6 +666,15 @@ def get_sqrtprec_from_sqrtprec(dim, sqrtprec, sparse_flag):
         logdet = np.sum(-np.log(sqrtprec.data**2))
         rank = dim
 
+    # sqrtprec is LinearOperator
+    elif isinstance(sqrtprec, spa.linalg.LinearOperator):
+        sqrtprec = sqrtprec
```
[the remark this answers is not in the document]
chaozg: Thanks for the spot. It's indeed unnecessary.

PR #424, opened 2024-05-26
Regularized gaussian extension

jeverink:
Summary of added features:
- Added ADMM solver
- Added Total Variation regularization for Regularized Gaussians
- Added RegularizedUniform implicit prior
- Added ConstrainedGaussian/GMRF and NonnegativeGaussian/GMRF as simplified alternatives
- Added Modified Half Normal distribution (conjugate distribution to RegularizedGaussian/GMRF)
- Added Gibbs support for l1 and Total Variation Regularization
- Added Gibbs sampler 'scan order' and 'initial guesses' to stabilize the Gibbs chains

Includes a notebook with various examples.

PR #429, opened 2024-05-30
Regularized Gaussian extension: ModifiedHalfNormal

### LINKED ISSUE #437: Regularized Gaussian extension: ModifiedHalfNormal
#429

jeverink:
closes #437 

Sub PR from #424 
- Added Modified Half Normal distribution (conjugate distribution to RegularizedGaussian/GMRF)

To-do:
- [x] Add tests

### DIFF
--- a/cuqi/distribution/_modifiedhalfnormal.py
+++ b/cuqi/distribution/_modifiedhalfnormal.py
@@ -1,0 +1,133 @@
+import numpy as np
+import scipy.stats as sps
+import scipy.special as special
+from cuqi.distribution import Distribution
+from cuqi.utilities import force_ndarray
+
+class ModifiedHalfNormal(Distribution):
+    """
+    Represents a modified half-normal (MHN) distribution, a three-parameter family of distributions generalizing the Gamma distribution.
+    The distribution is continuous with pdf 
+    f(x; alpha, beta, gamma) propto x^(alpha-1) * exp(-beta * x^2 + gamma * x)
+
+    The MHN generalizes the half-normal distribution, because
+    f(x; 1, beta, 0) propto exp(-beta * x^2)
+
+    The MHN generalizes the gamma distribution because
+    f(x; alpha, 0, -gamma) propto x^(alpha-1) * exp(- gamma * x)
+
+    Reference:
+    [1] Sun, et al. "The Modified-Half-Normal distribution: Properties and an efficient sampling scheme." Communications in Statistics-Theory and Methods
+    
+    Parameters
+    ----------
+    alpha : float
+        The polynomial exponent parameter of the MHN distribution. Must be positive.
+
+    beta : float
+        The quadratic exponential parameter of the MHN distribution. Must be positive.
+
+    gamma : float
+        The linear exponential parameter of the MHN distribution.
+
+    """
+    def __init__(self, alpha, beta, gamma, is_symmetric=False, **kwargs):
+        # Init from abstract distribution class
+        super().__init__(is_symmetric=is_symmetric,**kwargs) 
+
+        self.alpha = alpha
+        self.beta = beta
+        self.gamma = gamma
+
+    def logpdf(self, x): # Unnormalized
+        return (self.alpha - 1)*np.log(x) - self.beta * x * x + self.gamma * x
+
+
+    def _MHN_sample_gamma_proposal(self, alpha, beta, gamma, delta = None):
+        if delta is None:
+            delta = beta + (gamma*gamma - gamma*np.sqrt(gamma*gamma + 8*beta*alpha))/(4*alpha)
+            
+        while True:
+            T = np.random.gamma(alpha/2, 1.0/delta)
+            X = np.sqrt(T)
+            U = np.random.uniform()
+            if X > 0 and np.log(U) < -(beta-delta)*T + gamma*X - gamma*gamma/(4*(beta-delta)):
+                return X
+            
+    def _MHN_sample_normal_proposal(self, alpha, beta, gamma, mu):
+        if mu is None:
+            mu = (gamma + np.sqrt(gamma*gamma + 8*beta*(alpha - 1)))/(4*beta)    
+        
+        while True:
+            X = np.random.normal(mu, np.sqrt(0.5/beta))
+            U = np.random.uniform()
+            if X > 0 and np.log(U) < (alpha-1)*np.log(X) - np.log(mu) + (2*beta*mu-gamma)*(mu-X):
+                return X
+
+    # Sample from MHN when alpha > 1, beta > 0 and gamma > 0  (Algorithm 1 from [1])
+    def _MHN_sample_positive_gamma_1(self, alpha, beta, gamma):
+        if gamma <= 0.0:
+            raise ValueError("gamma needs to be positive")
+            
+        if alpha <= 1.0:
+            raise ValueError("alpha needs to be greater than 1.0")
+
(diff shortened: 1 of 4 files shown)


cuqi/distribution/_modifiedhalfnormal.py
```diff
@@ -0,0 +1,133 @@
...
+        The polynomial exponent parameter of the MHN distribution. Must be positive.
+
+    beta : float
+        The quadratic exponential parameter of the MHN distribution. Must be positive.
+
+    gamma : float
+        The linear exponential parameter of the MHN distribution.
+
+    """
+    def __init__(self, alpha, beta, gamma, is_symmetric=False, **kwargs):
```
nabriis, line 34:
```suggestion
    def __init__(self, alpha=None, beta=None, gamma=None, is_symmetric=False, **kwargs):
```
Distributions must be allowed to be initialized no arguments. Hopefully that fixes the unit test also.

cuqi/distribution/_modifiedhalfnormal.py
```diff
@@ -0,0 +1,133 @@
...
+        # Init from abstract distribution class
+        super().__init__(is_symmetric=is_symmetric,**kwargs)
```
nabriis, line 36:
```suggestion
        super().__init__(is_symmetric=is_symmetric, **kwargs) 

```

cuqi/distribution/_modifiedhalfnormal.py
```diff
@@ -0,0 +1,133 @@
...
+    def __init__(self, alpha, beta, gamma, is_symmetric=False, **kwargs):
+        # Init from abstract distribution class
+        super().__init__(is_symmetric=is_symmetric,**kwargs) 
+
+        self.alpha = alpha
+        self.beta = beta
+        self.gamma = gamma
+
+    def logpdf(self, x): # Unnormalized
+        return (self.alpha - 1)*np.log(x) - self.beta * x * x + self.gamma * x
```
nabriis, line 43: Looks like we could add the gradient also?

cuqi/distribution/_modifiedhalfnormal.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+    def _MHN_sample_gamma_proposal(self, alpha, beta, gamma, delta = None):
+        if delta is None:
```
nabriis, line 47: Please add docstring, even to private methods, that explain what the method does.

cuqi/distribution/_modifiedhalfnormal.py
```diff
@@ -0,0 +1,133 @@
...
+    def _MHN_sample(self, alpha, beta, gamma, m = None):
+        if gamma <= 0.0:
+            return self._MHN_sample_negative_gamma(alpha, beta, gamma, m = m)
+        
+        if alpha >= 1:
+            return self._MHN_sample_positive_gamma_1(alpha, beta, gamma)
+        
+        return self._MHN_sample_gamma_proposal(alpha, beta, gamma)
+
+    def _sample(self, N, rng=None):
```
nabriis, line 131: Looks like rng term is not getting used. It looks to me like the np.random is used to generate random numbers. Then you should be able to pass through a random number generator to the private methods and use that.

nabriis (changes requested):
Looks great Javerink! I had some comments here and there. Also It would be good if you could add some unit tests that regression test the sample and logpdf (and gradient) methods.

cuqi/distribution/_modifiedhalfnormal.py
jeverink: The gradient of the logpdf could be added.
I was wondering how important it is that the logpdf that I provided is unnormalized? Sadly, the normalization constant depends on a generalized generalized hypergeometric function, which might be relatively expensive to compute and have some numerical stability issues when implemented.

nabriis: It is OK to provide it unnormalized. We will soon update all distributions to use the unormalized in most cases.

nabriis: It would be fantastic if you could add the gradient @jeverink  :)

cuqi/distribution/_modifiedhalfnormal.py
```diff
@@ -0,0 +1,142 @@
...
+        The linear exponential parameter of the MHN distribution.
+
+    """
+    def __init__(self, alpha=None, beta=None, gamma=None, is_symmetric=False, **kwargs):
+        # Init from abstract distribution class
+        super().__init__(is_symmetric=is_symmetric, **kwargs) 
+
+        self.alpha = alpha
+        self.beta = beta
+        self.gamma = gamma
```
nabriis, line 40: Also. If you could add getters and setters for these 3 parameters with force_ndarray in them, then the unit tests that are skipped in test_distributions_shape should hopefully pass.

jeverink: I have added the requested getters and setters, with the force_ndarray, yet the tests still fail. With the Modified Half-Normal being a generalized Gamma distribution, I suppose the issues are the same as why these tests are skipped for the Gamma distribution and other related distributions. So if the issue is the same, shouldn't they all be fixed in one go, perhaps not as part of this pull request?

nabriis: To me it looks like the logd returns a vector. We generally assume that logd should return a scalar value. E.g. for multivariate Normal we sum the logd vector (see https://github.com/CUQI-DTU/CUQIpy/blob/b3c53f2ac21c807186dc35053421615c1cb3f205/cuqi/distribution/_normal.py#L40) Hopefully that helps?

jeverink: Thanks for the suggestion. I have fixed it. Another issue with the failed test is that is uses np.allclose(a, b), which returns False if both a and b are NaN, using np.allclose(a, b, equal_nan = True) will fix this issue. This happens when logpdf is evaluated outside its domain. With these changes, all tests now pass.

nabriis: Perfect. Thank you!

nabriis (changes requested):
Thank you @jeverink for the most recent changes. The ModifiedHalfNormal LGTM now. Only thing that I can see remains is to add some unit tests to test the logd, gradient and perhaps sample methods. I would suggest to add the following set of tests

1) Testing logd and gradient against known values/methods. See e.g. existing tests like `test_Cauchy_against_scipy` or `test_beta` in test_distribution.py. Perhaps here you can use the fact it is a generelizaiton of the Gamma distribution and test specific choices of the parameters that coinside with Gamma?

2) Add regression tests for the sampling. These can be similar to `test_Gaussian_sample_regression` in test_distribution.py. This is to ensure future changes to the code preserved the current behavior.

Once these two groups of tests have been added I am happy to get this PR merged. I will add @jakobsj as a second reviewer now to the PR.

jeverink:
Thanks for the suggestions @nabriis , I have now added a regression test for the sampling, logpdf and gradient.
Currently, due the the logpdf in the ModifiedHalfNormal being unmodified, the logpdf will not match with that of the Gamma. Also, there is no gradient implemented in Gamma currently, so such comparison is currently not possible.

nabriis (approved):
Thanks @jeverink. LGTM!

jakobsj (approved):
Looks great!  Thanks also for adding the unit tests!

PR #430, opened 2024-05-30
Regularized Gaussian extension: RegularizedUniform and Constraints/Nonnegative Gaussian/GMRF alternatives

### LINKED ISSUE #436: Regularized Gaussian extension: RegularizedUniform and Constraints/Nonnegative Gaussian/GMRF alternatives
#430

jeverink:
Closes #436 

Sub PR from #424 
- Added RegularizedUniform implicit prior
- Added ConstrainedGaussian/GMRF and NonnegativeGaussian/GMRF as simplified alternatives

Explanation of RegularizedUniform:
Consider a regularized Gaussian posterior of the form:
$\min_{x} \text{likelihood}(x, \hat{b}) + \frac12  \|\|x - \hat{\mu}\|\|_{\Sigma^{-1}}^2 + f(x)  $.
In various cases, we might not want the randomized linear least squares term from the RegularizedGaussian prior, but only the penalty $f(x)$. To make the randomized prior term disappear, we let $\Sigma^{-1} = 0$, or more precisely, $\Sigma^{-\frac{1}{2}} = 0$, such that the posterior takes the form:

$\min_{x} \text{likelihood}(x, \hat{b}) + f(x)  $.

This, the RegularizedUniform can be interpreted as a RegularizedGaussian with infinitely flat underlying Gaussian, i.e., a uniform "distribution" on the unbounded parameter space.

By letting the RegularizedUniform inherit from RegularizedGaussian, set the sqrtprec to zero, RegularizedLinearRTO will accept this implicit prior without modification. 

To-do:
- [x] Add tests


nabriis (changes requested):
Could you add some explanation on these additions in the PR?

I don't quite understand the Gaussian craeted with zeros as sqrtprec. Do we have some alternatives? Why are we not using the Uniform distribution?

jeverink:
Hi @nabriis ,
I've added an explanation as to why I have designed the RegularizedUniform class as such above, hope it helps.

cuqi/implicitprior/_regularizedUniform.py
```diff
@@ -0,0 +1,61 @@
+from cuqi.implicitprior import RegularizedGaussian
+from cuqi.distribution import Distribution, Gaussian
+
+import numpy as np
+
+
+class RegularizedUniform(RegularizedGaussian):
+        """ Implicit Regularized GMRF (Gaussian Markov Random Field).
```
jakobsj, line 8: I think docstring needs updating to RegularizedUniform

cuqi/implicitprior/_regularizedUniform.py
```diff
@@ -0,0 +1,61 @@
...
+                lower_bound : array_like or None
+                        Lower bound of box, defaults to zero
+                upper_bound : array_like
+                        Upper bound of box, defaults to one
+
+        regularization : string or None
+                Preset regularization. Can be set to "l1" or "TV". Required for use in Gibbs in future update.
+                For "l1", the following additional parameters can be passed:
+                strength : scalar
+                        Regularization parameter, i.e., strength*||x||_1 , defaults to one
```
jakobsj, line 45: How come this is only possible for l1 and not TV? Would expect to be able to set the regularization strength similarly for TV?

cuqi/implicitprior/_regularizedUniform.py
```diff
@@ -0,0 +1,61 @@
+from cuqi.implicitprior import RegularizedGaussian
+from cuqi.distribution import Distribution, Gaussian
+
+import numpy as np
+
+
+class RegularizedUniform(RegularizedGaussian):
```
jakobsj, line 7: I wonder if using "uniform" in the name can be misleading. A conventional uniform distribution will live only on a bounded domain with some finite value, whereas this lives on the entire space.  Would there be any sense in the name "RegularizedFlatGaussian" or even "RegularizedFlat". Here, "flat" replaces "uniform" to convey it is, well, flat, and detaches from the usual bounded domain associated with "uniform".  "gaussian" would keep the connection to the underlying nature of the class.  The second option "RegularizedFlat" is less clear and a bit odd, with the main attractive aspect being its shortness. Thoughts?

cuqi/implicitprior/_regularizedGMRF.py
```diff
@@ -68,3 +68,16 @@ def __init__(self, mean=None, prec=None, bc_type='zero', order=1, proximal = Non
             super(Distribution, self).__init__(**kwargs)
 
             self._parse_regularization_input_arguments(proximal, projector, constraint, regularization, args)
+
+
+
+class ConstrainedGMRF(RegularizedGMRF):
+    # TODO: Documentation
```
jakobsj, line 75: Please add documentation.

cuqi/implicitprior/_regularizedGMRF.py
```diff
@@ -68,3 +68,16 @@ def __init__(self, mean=None, prec=None, bc_type='zero', order=1, proximal = Non
...
+    def __init__(self, mean=None, prec=None, bc_type='zero', order=1, projector=None, constraint=None, **kwargs):
+        super().__init__(mean=mean, prec=prec, bc_type=bc_type, order=order, projector=projector, constraint=constraint, **kwargs)
+
+        
+class NonnegativeGMRF(RegularizedGMRF):
+    # TODO: Documentation
```
jakobsj, line 81: Please add.

cuqi/implicitprior/_regularizedGaussian.py
```diff
@@ -234,3 +234,15 @@ def _condition(self, *args, **kwargs):
             new_density = new_density.to_likelihood(value)
 
         return new_density
+
+
+class ConstrainedGaussian(RegularizedGaussian):
+    # TODO: Documentation
```
jakobsj, line 240: Please add.

cuqi/implicitprior/_regularizedGaussian.py
```diff
@@ -234,3 +234,15 @@ def _condition(self, *args, **kwargs):
...
+    def __init__(self, mean=None, cov=None, prec=None, sqrtcov=None,sqrtprec=None, projector=None, constraint=None, **kwargs):
+        super().__init__(mean=mean, cov=cov, prec=prec, sqrtcov=sqrtcov,sqrtprec=sqrtprec, projector=projector, constraint=constraint, **kwargs)
+
+        
+class NonnegativeGaussian(RegularizedGaussian):
+    # TODO: Documentation
```
jakobsj, line 246: Please add.

jakobsj (changes requested):
Thanks for this PR.  I have gone through on an overall design level.  I think it is useful to include these convenience "distributions" to simplify and more clearly communicate the availability of the common use cases of nonnegativity and simple constrained versions.  Namingwise, I think it works well, with the current suggestion, which keeps "Gaussian" ("GMRF") at the end, and replaces the "specifier" in front, e.g., "regularized" by "constrained" or "nonnegative", so I'm happy with that.  For the "uniform" one, I added some considerations.

Happy to discuss.

Please also include a couple of unit tests.

If you can address these comments first , that would be great.  I'll leave a further technical review to someone else afterwards.

jakobsj:
PS, should this PR remain a draft or be changed?

cuqi/implicitprior/_regularizedUniform.py
jeverink: My two kroner on the matter:

As for "RegularizedFlatGaussian":
 The reason the current RegularizedUniform class inherits from RegularizedGaussian is to make RegularizedLinearRTO work without any modifications. However, this is far from optimal as computations are being done with an all zero matrix (the sqrtprec). It could be that the RegularizedLinearRTO and the current RegularizedUniform might be reworked in the future to not require this anymore, resulting in there no longer being any underlying Gaussian. It is with that argument in mind that I prefer not to refer to this implicit prior with the Gaussian suffix.

Regarding "RegularizedUniform" or "RegularizedFlat":
 I agree that a uniform distribution can only _properly_ exist on a bounded domain, however, within the Bayesian literature, there is distinction between _proper_ and _improper_ priors, based on whether the prior represents an actual probability distribution or not. A uniform prior on a unbounded domain (or more precisely on a domain of infinite measure) is a commonly used example of such an improper prior. In CUQIpy, as far as I am aware, you can create such improper priors already, for example, by creating a Gaussian "distribution" with a sqrtprec that is not full-rank. This is precisely the trick I currently use to make the RegularizedUniform class work. It is wat that argument in mind that I prefer to refer to this implicit prior with the Uniform suffix.

I hope this clarifies my choice for the naming decision a little bit, and perhaps spark a little bit of a discussion on CUQIpy conventions for proper and improper priors.

cuqi/implicitprior/_regularizedUniform.py
jeverink: Good catch, this is a remnant from me copying code from a version that has TV implemented. In this PR TV is not yet available, so I remove this when updating the docstring.

jeverink:
Thanks for the comment @jakobsj ,
I have now updated all the documentation and added some tests.

cuqi/implicitprior/_regularizedUniform.py
nabriis: How about RegularizedUnboundedUniform?

jeverink: I am fine with that, how about you @jakobsj ?

jakobsj: Many thanks for the background. I agree RegularizedUnboundedUniform is a sensible name.

jakobsj (approved):
Thanks for adding tests and documentation, it looks good to me. I'm approving now, so if you just update the name as discussed, then we can pass it to @nabriis for a last technical look.

jeverink:
I have now renamed RegularizedUniform to RegularizedUnboundedUniform

jakobsj:
Thanks @jeverink. I've rerequested review from Nicolai.

cuqi/implicitprior/_regularizedGaussian.py
```diff
@@ -234,3 +234,90 @@ def _condition(self, *args, **kwargs):
...
+
+    sqrtcov
+        See :class:`~cuqi.distribution.Gaussian` for details.
+
+    sqrtprec
+        See :class:`~cuqi.distribution.Gaussian` for details.
+
+    """
+    def __init__(self, mean=None, cov=None, prec=None, sqrtcov=None,sqrtprec=None, **kwargs):
+        super().__init__(mean=mean, cov=cov, prec=prec, sqrtcov=sqrtcov,sqrtprec=sqrtprec, constraint="nonnegativity", **kwargs)
```
nabriis, line 323:
```suggestion
        super().__init__(mean=mean, cov=cov, prec=prec, sqrtcov=sqrtcov, sqrtprec=sqrtprec, constraint="nonnegativity", **kwargs)
```

cuqi/implicitprior/_regularizedGaussian.py
```diff
@@ -234,3 +234,90 @@ def _condition(self, *args, **kwargs):
...
+        Preset constraints. Can be set to "nonnegativity" and "box". Required for use in Gibbs.
+        For "box", the following additional parameters can be passed:
+            lower_bound : array_like or None
+                Lower bound of box, defaults to zero
+            upper_bound : array_like
+                Upper bound of box, defaults to one
+
+    """
+    def __init__(self, mean=None, cov=None, prec=None, sqrtcov=None,sqrtprec=None, projector=None, constraint=None, **kwargs):
+        super().__init__(mean=mean, cov=cov, prec=prec, sqrtcov=sqrtcov,sqrtprec=sqrtprec, projector=projector, constraint=constraint, **kwargs)
```
nabriis, line 287:
```suggestion
        super().__init__(mean=mean, cov=cov, prec=prec, sqrtcov=sqrtcov, sqrtprec=sqrtprec, projector=projector, constraint=constraint, **kwargs)
```

cuqi/implicitprior/_regularizedUnboundedUniform.py
```diff
@@ -0,0 +1,66 @@
...
+
+        For more details on implicit regularized Gaussian see the following paper:
+
+        [1] Everink, Jasper M., Yiqiu Dong, and Martin S. Andersen. "Sparse Bayesian inference with regularized
+        Gaussian distributions." Inverse Problems 39.11 (2023): 115004.
+
+        Parameters
+        ----------
+        geometry : :class:`~cuqi.geometry.Geometry`
+                The geometry of the underlying variable.
```
nabriis, line 28:
```suggestion
        geometry : :class:`~cuqi.geometry.Geometry` or integer
                The geometry of the underlying variable. Defines the dimension of the distribution. 
```

cuqi/implicitprior/_regularizedUnboundedUniform.py
```diff
@@ -0,0 +1,66 @@
+from cuqi.implicitprior import RegularizedGaussian
+from cuqi.distribution import Distribution, Gaussian
+
+import numpy as np
+
+
+class RegularizedUnboundedUniform(RegularizedGaussian):
```
nabriis, line 7: I was wondering if you had a quick demo you could throw together to illustrate the use of this  implicit prior?

tests/test_implicit_priors.py
```diff
@@ -28,3 +28,43 @@ def test_RegularizedGaussian_guarding_statements():
...
+
+def test_NonnegativeGMRF_alias():
+    """ Test that the implicit nonnegative GMRF is a correct allias for an implicit regularized GMRF """
+
+    x = cuqi.implicitprior.NonnegativeGMRF(np.zeros(5), 1)
+
+    assert isinstance(x, cuqi.implicitprior.RegularizedGMRF)
+    assert x.preset == "nonnegativity"
+
+def test_RegularizedUnboundedUniform():
```
nabriis, line 64:
```suggestion
def test_RegularizedUnboundedUniform_is_RegularizedGaussian():
```

nabriis (changes requested):
Thanks @jeverink. I like the "alias" or easy to get to distributions. I had a few small suggestions and a request for a short demo on the unbounded uniform.

Thanks for adding the tests. They are great and much appreciated!

cuqi/implicitprior/_regularizedUnboundedUniform.py
jeverink: I have added a quick demo notebook that shows it, and some of the others, in action.

nabriis: Thanks @jeverink. LGTM!

nabriis (approved):

PR #446, opened 2024-06-28
Find valid samplers

### LINKED ISSUE #447: review PR on find_valid_samplers
review #446

jeverink:
Added a method that returns a list of all samplers that should pass validate_target.

Also fixes a bug where the two proposal based samplers are valid for implicit distribution.
This fix assumes that implicit distributions return NaN when logd is called, which is currently the case.
(Charlie edited: the above is for correctly identifying `CWMH` and `MH` as __invalid__ for targets with implicit priors)

Note: I have not checked whether the whether the returned samplers to actually work, so the added tests might not be proper.

closed #447

### DIFF
--- a/cuqi/experimental/mcmc/__init__.py
+++ b/cuqi/experimental/mcmc/__init__.py
@@ -1,6 +1,6 @@
 """ Re-implementation of sampler module in a more object oriented way. """
 
-from ._sampler import SamplerNew, ProposalBasedSamplerNew, find_valid_samplers
+from ._sampler import SamplerNew, ProposalBasedSamplerNew
 from ._langevin_algorithm import ULANew, MALANew
 from ._mh import MHNew
 from ._pcn import PCNNew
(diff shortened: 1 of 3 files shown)


chaozg:
This is excellent! @jeverink

nabriis:
Looks great @jeverink. Will discuss with team on where to put the utility method and its name. But fantastic idea!

chaozg (changes requested):
Hi @jeverink , this PR will definitely improve a user's experience with CUQIpy and I'm very happy to see such a neat implementation and so extensive test cases. After a brief group discussion with @jakobsj and @nabriis , the only suggestion I have so far is to put `find_valid_samplers ` to a more proper place, like `cuqi/experimental/mcmc/_utilities.py`.

chaozg (approved):

jakobsj (approved):
Great idea and elegant solution! Thanks very much for contributing this utility. Already in great shape so I see no need for any changes.

PR #449, opened 2024-07-01
HMC sampler

furibec:
- The file `_hmc.py` contained the NUTS implementation, to I renamed it to `_nuts.py`
- Fixed a typo in the kinetic energy function of NUTS: it was `0.5*(r.T @ r)`, but I think it must be `-0.5*(r.T @ r)` (the actual density value not its negative).
- Created the actual HMC sampler.
- Test example to be done to see if the implementation is correct.

### DIFF
--- a/cuqi/sampler/_hmc.py
+++ b/cuqi/sampler/_hmc.py
@@ -2,25 +2,28 @@ import numpy as np
 from cuqi.sampler import Sampler
 
 
-# another implementation is in https://github.com/mfouesneau/NUTS
-class NUTS(Sampler):
-    """No-U-Turn Sampler (Hoffman and Gelman, 2014).
+class HMC(Sampler):
+    """Hamiltonian Monte Carlo (Duane et al., 1987).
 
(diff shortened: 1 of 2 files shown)


furibec:
These are the step sizes (epsilon) results for NUTS and HMC. Notice that in HMC, after warm-up, the step sizes keep changing. This is because we have to do a jittering of the epsilons in order to avoid pathological behavior
image

furibec:
and this is the result for the main statistics estimated by NUTS and HMC. Here we are sampling a 100-dimensional Gaussian with zero mean and diagonal covariance. The samples are both estimating very well the reference mean and standard deviations.
image

furibec:
I believe the samplers are working as expected and they are also in CUQIpy format. Nevertheless, some test are failing due to the changes I did in NUTS. I hope someone can take it from here @jakobsj

jakobsj:
Thank you very much @furibec for adding your HMC sampler, the fixes as well as a demo validating the implementation, that is most helpful!  I cannot promise when we'll be able to look at this but I hope it will be soon.

PR #450, opened 2024-07-01
Add SmoothedLaplace

### LINKED ISSUE #451: add SmoothedLaplace

chaozg:
fixed #451 

This PR adds SmoothedLaplace, an approximation of Laplace, and its pdf is evaluated as 
$$p(x) = \frac{1}{2b} \exp\left(-\frac{\sqrt{(x-\mu)^2 + \beta}}{b}\right),$$
where $\beta$ is a smoothing parameter.

Compared with Laplace, SmoothedLaplace comes with a `gradient` method and it enables the use of gradient-based samplers.

Smoothed Laplace was implemented earlier as a custom distribution in showcase notebooks on [smoothed Laplace](https://github.com/CUQI-DTU/CUQIpy-User-Showcase/blob/main/007_smoothed_Laplace/smoothed_laplace.ipynb) and [super resolution](https://github.com/CUQI-DTU/CUQIpy-User-Showcase/blob/main/012_super_resolution/STORM.ipynb).

### DIFF
--- a/cuqi/distribution/_smoothed_laplace.py
+++ b/cuqi/distribution/_smoothed_laplace.py
@@ -1,4 +1,5 @@
 import numpy as np
+from cuqi.utilities import force_ndarray
 from cuqi.distribution import Distribution
 
 class SmoothedLaplace(Distribution):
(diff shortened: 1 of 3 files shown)


jakobsj (approved):
Brilliant! Thanks for adding this from the showcase into the main library!

cuqi/distribution/_smoothed_laplace.py
```diff
@@ -0,0 +1,80 @@
...
+        The location parameter of the distribution.
+
+    scale : scalar, list, tuple, or ndarray
+        The scale parameter of the distribution.
+    
+    beta : scalar
+
+    """
+
+    def __init__(self, location, scale, beta=1e-3, **kwargs):
```
nabriis, line 33: Can you ensure the variables location and scale can be set to None? This is needed for conditioning, Gibbs etc.
It can be achieved using getters and setters. See e.g. Cauchy.py for an example that is somewhat similar.
The mutable variables for the distribution here should return location, scale.

tests/test_distribution.py
```diff
@@ -779,4 +779,30 @@ def test_MHN_regression(alpha, beta, gamma, expected_logpdf, expected_gradient):
...
-    assert np.allclose( gradient, np.array(expected_gradient))
\ No newline at end of file
+    assert np.allclose( gradient, np.array(expected_gradient))
+
+def test_Smoothed_Laplace():
+    """ Test Smoothed Laplace distribution logpdf and gradient """
+
+    location = np.array([1, 2])
+    scale = np.array([1, 2])
+    scaller_laplace_0 = cuqi.distribution.Laplace(location[0], scale[0])
```
nabriis, line 789: Scalar?

nabriis (changes requested):
Hi @chaozg. Looks great! I had some minor comments.

cuqi/distribution/_smoothed_laplace.py
nabriis: Also add defaults location=None, scale=None please :)

tests/test_distribution.py
chaozg: Done

cuqi/distribution/_smoothed_laplace.py
chaozg: Done as suggested. Thanks!

PR #472, opened 2024-08-11
Add repr method for cuqi samples

### LINKED ISSUE #468: Add method to display easy-to-read string for Samples object
**Description**

when calling the function print on the array .sample(n>1) the output is
 `cuqi.samples._samples.Samples object at 0x0000013513CFA3C0`

It would be more useful to display some meaningful information, such as its dimension.

chaozg:
- fixed #468

Now when we `evaluate/print` a cuqi Samples, it will display something like:
```bash
CUQIpy Samples:
---------------

Ns (number of samples):
 2

Geometry:
 _DefaultGeometry1D(4,)

Shape:
 (4, 2)

Samples:
 [[ 1.38560564  0.53052923]
 [ 0.4737199   0.51313651]
 [-0.20398077 -1.29972158]
 [-2.48091298  0.27251739]]
```

### DIFF
--- a/cuqi/geometry/_geometry.py
+++ b/cuqi/geometry/_geometry.py
@@ -2,7 +2,6 @@ from abc import ABC, abstractmethod
 import numpy as np
 import numpy.matlib as matlib
 import matplotlib.pyplot as plt
-from matplotlib.ticker import MaxNLocator
 import math
 from scipy.fftpack import dst, idst
 import scipy.sparse as sparse
--- a/cuqi/geometry/_geometry.py
+++ b/cuqi/geometry/_geometry.py
@@ -410,6 +410,7 @@ class Continuous1D(Continuous):
         if self.axis_labels is not None:
             plt.xlabel(self.axis_labels[0])
 
+
 class Continuous2D(Continuous):
 
     def __init__(self,grid=None,axis_labels=None):
--- a/cuqi/geometry/_geometry.py
+++ b/cuqi/geometry/_geometry.py
@@ -410,8 +410,6 @@ class Continuous1D(Continuous):
     def _plot_config(self):
         if self.axis_labels is not None:
             plt.xlabel(self.axis_labels[0])
-        plt.gca().xaxis.set_major_locator(MaxNLocator(integer=True))
-
 
 class Continuous2D(Continuous):
 
--- a/cuqi/geometry/_geometry.py
+++ b/cuqi/geometry/_geometry.py
@@ -2,6 +2,7 @@ from abc import ABC, abstractmethod
 import numpy as np
 import numpy.matlib as matlib
 import matplotlib.pyplot as plt
+from matplotlib.ticker import MaxNLocator
 import math
 from scipy.fftpack import dst, idst
 import scipy.sparse as sparse
@@ -409,6 +410,7 @@ class Continuous1D(Continuous):
     def _plot_config(self):
         if self.axis_labels is not None:
             plt.xlabel(self.axis_labels[0])
+        plt.gca().xaxis.set_major_locator(MaxNLocator(integer=True))
 
 
 class Continuous2D(Continuous):
--- a/cuqi/samples/_samples.py
+++ b/cuqi/samples/_samples.py
@@ -874,8 +874,6 @@ class Samples(object):
 
         return ax
 
-    def __repr__(self) -> str:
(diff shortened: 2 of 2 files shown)


cuqi/samples/_samples.py
```diff
@@ -873,3 +873,6 @@ def plot_violin(self, variable_indices=None, **kwargs):
         ax =  arviz.plot_violin(datadict, **kwargs)
 
         return ax
+
+    def __repr__(self) -> str:
+        return "A CUQIpy samples object with shape {}".format(self.samples.shape)
```
amal-ghamdi, line 878:
```suggestion
        return "A CUQIpy Samples object with shape {}".format(self.samples.shape)
```

cuqi/samples/_samples.py
```diff
@@ -873,3 +873,6 @@ def plot_violin(self, variable_indices=None, **kwargs):
         ax =  arviz.plot_violin(datadict, **kwargs)
 
         return ax
+
+    def __repr__(self) -> str:
```
amal-ghamdi, line 877: Displaying the shape is a really nice addition, I suggest we can take this a step further to mention which dimension is the number of samples and which is the dim of the samples. Hence, the properties: `Ns` for the number of samples and `s.geometry.par_dim` for the dimension of samples can be used instead of the shape

amal-ghamdi (changes requested):
Thank you @chaozg for this nice addition :). I made a small comment here.

cuqi/samples/_samples.py
chaozg: Done

cuqi/samples/_samples.py
chaozg: Done

cuqi/geometry/_geometry.py
```diff
@@ -409,6 +410,7 @@ def _plot_envelope(self, lo_values, up_values, **kwargs):
     def _plot_config(self):
         if self.axis_labels is not None:
             plt.xlabel(self.axis_labels[0])
+        plt.gca().xaxis.set_major_locator(MaxNLocator(integer=True))
```
nabriis, line 413: Since this belongs to Continous1D, I find it a bit strange to restrict tick locations to integers only.  Perhaps a more robust solution is to update the behavior of the default Geometry - which I believe is the cause of the described problem in the issue - by e.g. subclassing from a more generic class than Continous1D. Perhaps we can make a default Geometry class that has certain behavior when # of x-valeus are low, and another for higher.

cuqi/samples/_samples.py
nabriis: While we are at it, why not list the geometry?

cuqi/geometry/_geometry.py
amal-ghamdi: I agree here with @nabriis that setting the ticks to integer might not be the best for `Continuous1D`. One possibility is to override `_plot_config` in  `_DefaultGeometry` to have this line (line 413), and maybe set the display to marker (dots or o's) rather than lines. Or make the `_DefaultGeometry` a subclass of discrete instead of `Continuous1D`. I am not sure though what is the best solution.

You can make this another pull request if you like so the Samples __repr__ can be merged soon.

amal-ghamdi (changes requested):
Thank you @chaozg for the update. I have a small comment regarding the default geometry

cuqi/samples/_samples.py
chaozg: Done

cuqi/geometry/_geometry.py
chaozg: Thank you both for your guidance and thoughts here. I now understand my update at Continuous1D didn't make sense at all and the existing methods to handle plot is already very good. I quote from @nabriis , "There is not great solution without being explicit and setting geometry manually". So, I'll withdraw my naive change here and leave the choice to the user.

amal-ghamdi (approved):
Thank you @chaozg, I like the updated print of the `Samples` object. And it is great that it is consistent with the CUQIarray print. Approved :)

cuqi/geometry/_geometry.py
nabriis: Ok let us leave it like it is then.

nabriis (approved):
LGTM

PR #482, opened 2024-08-20
Ensure constants are stored in densities when reducing to single density in joint distribution

### LINKED ISSUE #481: Joint distribution conditioning on out-of-bounds values
**Description**

When working with joint distributions and evaluating log density or conditioning, we need to ensure the log density is consistant when evaluating. 

This is currently NOT the case if we condition on a joint distribution where some densities have the possibility of returning "-inf" when the value is out-of-bounds.

**Example to reproduce**

TBD

```python
import numpy as np
import cuqi

joint = cuqi.distribution.JointDistribution(
            cuqi.distribution.Uniform(0, 10, name="d"),
            cuqi.distribution.Uniform(0, 5, name="s"),
            cuqi.distribution.Gaussian(np.zeros(8), lambda d: d, name="x"),
            cuqi.distribution.Gaussian(
                mean=cuqi.testproblem.Deconvolution1D(dim=8).model,
                cov=lambda s: s,
                name="y"
            )
        )

print(joint.logd(d=11, s=2, x=np.zeros(8), y=cuqi.testproblem.Deconvolution1D(dim=8).data))

print(joint(d=10, s=2, y=cuqi.testproblem.Deconvolution1D(dim=8).data).logd(x=np.zeros(8)))
```

returns
```bash
[-inf]
[-30.61073851]
```
should return both [-inf]

nabriis:
Closes #481

### DIFF
--- a/cuqi/distribution/_joint_distribution.py
+++ b/cuqi/distribution/_joint_distribution.py
@@ -187,7 +187,6 @@ class JointDistribution:
         This method is a hack to allow our current samplers to work with
         the joint distribution. It should be removed in the future.
         """
-
         # Count number of distributions and likelihoods
         n_dist = len(self._distributions)
         n_likelihood = len(self._likelihoods)
@@ -214,14 +213,18 @@ class JointDistribution:
         # If no distributions and exactly one likelihood its a Likelihood
         if n_likelihood == 1 and n_dist == 0:
             return self._likelihoods[0]
+
+        # If only evaluated densities left return joint to ensure logd method is available
+        if n_dist == 0 and n_likelihood == 0:
+            return self
         
     def _add_constants_to_density(self, density: Density):
         """ Add the constants (evaluated densities) to a single density. Used when reducing to single density. """
-        constant = self._sum_evaluated_densities()
+
         if isinstance(density, EvaluatedDensity):
-            density.value += constant
-        else:
-            density._constant += constant
+            raise ValueError("Cannot add the sum of all evaluated densities to an EvaluatedDensity.")
+
+        density._constant += self._sum_evaluated_densities()
         return density
 
     def _as_stacked(self) -> _StackedJointDistribution:
--- a/cuqi/distribution/_joint_distribution.py
+++ b/cuqi/distribution/_joint_distribution.py
@@ -175,6 +175,10 @@ class JointDistribution:
                 kwargs[ordered_keys[index]] = arg
         return kwargs
 
+    def _sum_evaluated_densities(self):
+        """ Return the sum of the evaluated densities in the joint distribution """
+        return sum([density.logd() for density in self._evaluated_densities])
+
     def _reduce_to_single_density(self):
         """ Reduce the joint distribution to a single density if possible.
 
@@ -201,7 +205,10 @@ class JointDistribution:
             # Ensure parameter names match, otherwise return the joint distribution
             if set(self._likelihoods[0].get_parameter_names()) != set(self._distributions[0].get_parameter_names()):
                 return self
-            return Posterior(self._likelihoods[0], self._distributions[0])
+            # If match, create posterior and set constant to the sum of the EvaluatedDensities
+            posterior = Posterior(self._likelihoods[0], self._distributions[0])
+            posterior._constant = self._sum_evaluated_densities()
+            return posterior
         
         # If exactly one distribution and no likelihoods its a Distribution
         if n_dist == 1 and n_likelihood == 0:
--- a/cuqi/distribution/_joint_distribution.py
+++ b/cuqi/distribution/_joint_distribution.py
@@ -205,18 +205,24 @@ class JointDistribution:
             # Ensure parameter names match, otherwise return the joint distribution
             if set(self._likelihoods[0].get_parameter_names()) != set(self._distributions[0].get_parameter_names()):
                 return self
-            # If match, create posterior and set constant to the sum of the EvaluatedDensities
-            posterior = Posterior(self._likelihoods[0], self._distributions[0])
-            posterior._constant = self._sum_evaluated_densities()
-            return posterior
-        
+            return self._add_constants_to_density(Posterior(self._likelihoods[0], self._distributions[0]))
+
         # If exactly one distribution and no likelihoods its a Distribution
         if n_dist == 1 and n_likelihood == 0:
-            return self._distributions[0]
-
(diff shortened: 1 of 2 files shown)


tests/test_joint_distribution.py
```diff
@@ -310,3 +310,86 @@ def test_MultipleLikelihoodPosterior_should_raise_if_names_do_not_match():
...
+    ]
+)
+def test_logd_consistency_when_conditioning(joint, variables):
+    """ Test consistency of logd value when conditioning the joint distribution.
+     
+    This ensures we always return the correct value for logd even when reducing to single density.
+    
+    """
+
+    # True value og logd by fully evaluating the joint distribution
```
chaozg, line 369:
```suggestion
    # True value of logd by fully evaluating the joint distribution
```

tests/test_joint_distribution.py
```diff
@@ -310,3 +310,86 @@ def test_MultipleLikelihoodPosterior_should_raise_if_names_do_not_match():
...
+        # Condition the joint distribution
+        cond_joint = joint(**cond_vars)
+
+        # Evaluate the logd value of conditioned joint distribution
+        # This may be a single density or a joint distribution since
+        # joint can reduce to single density (such as Posterior)
+        logd_value = cond_joint.logd(**remaining_vars)
+
+        # Potential error message if this assert fails:
+        msg = f"Evaluated: \n {joint} \n with variables {variables}.\nFailed at {key}={value} for \n {cond_joint} \n with variables {remaining_vars}."
```
chaozg, line 389: Just a tiny reminder: this line of code is quite long and the display might also go quite long, for example:
```bash
with variables {'d': 5, 's': 2, 'x': array([0., 0., 0., 0., 0., 0., 0., 0.]), 'y': CUQIarray: NumPy array wrapped with geometry.
```

chaozg (commented):
Thanks @nabriis for this nice PR and the accompanying test cases. I only leave some tiny comments.

tests/test_joint_distribution.py
nabriis: Yes thanks. Its for the unit tests. I think it is OK for now.

chaozg (approved):
LGTM!

cuqi/distribution/_joint_distribution.py
```diff
@@ -201,15 +205,24 @@ def _reduce_to_single_density(self):
...
+        
         # If no distributions and exactly one likelihood its a Likelihood
         if n_likelihood == 1 and n_dist == 0:
             return self._likelihoods[0]
+        
+    def _add_constants_to_density(self, density: Density):
+        """ Add the constants (evaluated densities) to a single density. Used when reducing to single density. """
+        constant = self._sum_evaluated_densities()
+        if isinstance(density, EvaluatedDensity):
+            density.value += constant
```
amal-ghamdi, line 222: just to double-check here, would that add the constant of the evaluated density twice? could we just do `density.value = constant`? without `+=`. 

Not sure though

cuqi/distribution/_joint_distribution.py
```diff
@@ -175,6 +175,10 @@ def _parse_args_add_to_kwargs(self, *args, **kwargs):
                 kwargs[ordered_keys[index]] = arg
         return kwargs
 
+    def _sum_evaluated_densities(self):
+        """ Return the sum of the evaluated densities in the joint distribution """
+        return sum([density.logd() for density in self._evaluated_densities])
+
     def _reduce_to_single_density(self):
```
amal-ghamdi, line 182: Do we also need to update the case when we have `MultipleLikelihoodPosterior` to add constants?

amal-ghamdi (commented):
Hi @nabriis, this looks great and very nice testing! I added a few questions.

cuqi/distribution/_joint_distribution.py
nabriis: No. It is because the EvaluatedDensity is a bit a special case. It essentially represents a constant when e.g. we evaluate a logd using all variables. It uses the attribute .value to store the constant rather than constant. Perhaps we can update that in another PR if we want.

cuqi/distribution/_joint_distribution.py
nabriis: No because MultipleLikelihoodPosterior subclasses from JointDistribution. Therefore it knows and will potentially have EvaluatedDensities. Note the line ``return MultipleLikelihoodPosterior(*self._densities)`` all densities are returned. This is different than the single density cases, where we dont have a datastructure to store more than e.g. prior and likelihood.

nabriis:
Thanks for the review @amal-ghamdi. I answered your questions. Let me know if there are any further or if my answer does not make sense :)

cuqi/distribution/_joint_distribution.py
amal-ghamdi: Thank you @nabriis, I meant here the `+=` instead of `=`. Because it seemed to me that self._sum_evaluated_densities() will already count the constant in `density.value`, so it seemed to me we are incorporating it twice. But still not sure

cuqi/distribution/_joint_distribution.py
amal-ghamdi: got it! thank you for the clarification

amal-ghamdi (commented):
Thank you @nabriis for the clarification :). In one of the questions I have a follow-up comment.

cuqi/distribution/_joint_distribution.py
nabriis: This was a good catch @amal-ghamdi. This code path was actually never used, but may potentially have been in the future. I added a guarding statement to avoid accidentally attempting to add constants to the evaluated density in the future. Your question also made my find the unused case where one does as follows:
```python
import cuqi
d = cuqi.distribution.Gamma(1, 1)
s = cuqi.distribution.Gamma(1, 1)
x = cuqi.distribution.Gaussian(0, 1)
J = cuqi.distribution.JointDistribution(d,s,x)
J(d=1, s=1, x=1)
```
This currently returns "None", but should return sometihng that represents a constant. I fixed that also.

amal-ghamdi (approved):
Thank you @nabriis for addressing this. LGTM!

cuqi/distribution/_joint_distribution.py
amal-ghamdi: 👍

PR #484, opened 2024-08-22
Fix HybridGibbs using initial point from sampling strategy

### LINKED ISSUE #473: HybridGibbs initial point handling

It should be possible to set initial point as part of the sampling strategy e.g
```python
# Gibbs sampler
sampling_strategy = {
    "cx0" : MHNew(scale = sample_scale, inititial_point=0.3),
    "cx1" : MHNew(scale = sample_scale, inititial_point=0.5),
}
sampler = HybridGibbsNew(posterior, sampling_strategy)
```
and have the Gibbs sampler use that.

nabriis:
Closes #473 

Had to update how we create default initial point to not assume we have the dimension of the target at point of creation.

### DIFF
--- a/cuqi/experimental/mcmc/_gibbs.py
+++ b/cuqi/experimental/mcmc/_gibbs.py
@@ -255,8 +255,8 @@ class HybridGibbs:
         """ Get initial points for each parameter """
         initial_points = {}
         for par_name in self.par_names:
-            sampler = samplers[par_name]
-            if not sampler.initial_point:
+            sampler = self.samplers[par_name]
+            if sampler.initial_point is None:
                 sampler.initial_point = sampler._get_default_initial_point(self.target.get_density(par_name).dim)
             initial_points[par_name] = sampler.initial_point
             
--- a/cuqi/experimental/mcmc/_gibbs.py
+++ b/cuqi/experimental/mcmc/_gibbs.py
@@ -255,10 +255,11 @@ class HybridGibbs:
         """ Get initial points for each parameter """
         initial_points = {}
         for par_name in self.par_names:
-            if hasattr(self.target.get_density(par_name), 'init_point'):
-                initial_points[par_name] = self.target.get_density(par_name).init_point
-            else:
-                initial_points[par_name] = np.ones(self.target.get_density(par_name).dim)
+            sampler = samplers[par_name]
+            if not sampler.initial_point:
+                sampler.initial_point = sampler._get_default_initial_point(self.target.get_density(par_name).dim)
+            initial_points[par_name] = sampler.initial_point
+            
         return initial_points
 
     def _store_samples(self):
--- a/tests/zexperimental/test_mcmc.py
+++ b/tests/zexperimental/test_mcmc.py
@@ -812,7 +812,7 @@ def test_HybridGibbs_initial_point_setting():
     # Sampling strategy
     sampling_strategy = {
         "d" : cuqi.experimental.mcmc.MH(initial_point=3),
-        "s" : cuqi.experimental.mcmc.MH(initial_point=2),
+        "s" : cuqi.experimental.mcmc.MH(),
         "x" : cuqi.experimental.mcmc.MALA(initial_point=0.5*np.ones(10))
     }
 
@@ -821,6 +821,5 @@ def test_HybridGibbs_initial_point_setting():
 
     # Test that the initial point is set correctly in Gibbs
     assert sampler.current_samples["d"] == 3
-    assert sampler.current_samples["s"] == 2
+    assert sampler.current_samples["s"] == 1 # Default initial point of MH is 1.
     assert np.allclose(sampler.current_samples["x"], 0.5*np.ones(10))
-        
--- a/tests/zexperimental/test_mcmc.py
+++ b/tests/zexperimental/test_mcmc.py
@@ -788,3 +788,39 @@ def test_find_valid_samplers_implicit_prior():
     valid_samplers = cuqi.experimental.mcmc.find_valid_samplers(target)
 
     assert(len(set(valid_samplers)) == 0)
+
+# ============ Testing of HybridGibbs ============
+
+def test_HybridGibbs_initial_point_setting():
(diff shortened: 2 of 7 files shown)


cuqi/experimental/mcmc/_gibbs.py
```diff
@@ -255,10 +255,11 @@ def _get_initial_points(self):
         """ Get initial points for each parameter """
         initial_points = {}
         for par_name in self.par_names:
-            if hasattr(self.target.get_density(par_name), 'init_point'):
-                initial_points[par_name] = self.target.get_density(par_name).init_point
```
amal-ghamdi, line 259: I wonder if we have to make an issue to remove the initial point from distributions in the future. Seems unintuitive to me. (by in the future I mean after removing the old Sampler module)

cuqi/experimental/mcmc/_gibbs.py
```diff
@@ -255,10 +255,11 @@ def _get_initial_points(self):
...
-            else:
-                initial_points[par_name] = np.ones(self.target.get_density(par_name).dim)
+            sampler = self.samplers[par_name]
+            if sampler.initial_point is None:
+                sampler.initial_point = sampler._get_default_initial_point(self.target.get_density(par_name).dim)
+            initial_points[par_name] = sampler.initial_point
```
amal-ghamdi, line 261: looks much nicer :)

tests/zexperimental/test_mcmc.py
```diff
@@ -788,3 +788,38 @@ def test_find_valid_samplers_implicit_prior():
...
+    # Posterior
+    posterior = joint(y=y_data)
+
+    # Sampling strategy
+    sampling_strategy = {
+        "d" : cuqi.experimental.mcmc.MH(initial_point=3),
+        "s" : cuqi.experimental.mcmc.MH(),
+        "x" : cuqi.experimental.mcmc.MALA(initial_point=0.5*np.ones(10))
+    }
+
```
amal-ghamdi, line 818: Comment not very much to this PR. Do you think we need an issue to capture and prevent the case if the user created an MH object and passed it to both "d" and "s". For example, this can be done by ensuring all the sampler objects passed to Gibbs are distinct.

tests/zexperimental/test_mcmc.py
```diff
@@ -788,3 +788,38 @@ def test_find_valid_samplers_implicit_prior():
...
+    # Hybrid Gibbs sampler
+    sampler = cuqi.experimental.mcmc.HybridGibbs(posterior, sampling_strategy=sampling_strategy)
+
+    # Test that the initial point is set correctly in Gibbs
+    assert sampler.current_samples["d"] == 3
+    assert sampler.current_samples["s"] == 1 # Default initial point of MH is 1.
+    assert np.allclose(sampler.current_samples["x"], 0.5*np.ones(10))
```
amal-ghamdi, line 825: do we need to also sample and assert first sample is the initial point, or is it redundant?

amal-ghamdi (changes requested):
Hi @nabriis, this looks like a very nice solution for the initialization issue! I have a few comments, nothing major. most of them are question-like.

cuqi/experimental/mcmc/_gibbs.py
nabriis: Actually they were not part of the distributions at all. If was this "hack" that one had to add an attribute to an instance of a distribution. So this removes it entirely.

tests/zexperimental/test_mcmc.py
nabriis: Good point. #485

tests/zexperimental/test_mcmc.py
nabriis: Not needed. In fact first sample will not be the initial point in the new samplers. But we need to ensure that the samplers current point after initialization starts at the initial point.

amal-ghamdi: got it!

cuqi/experimental/mcmc/_gibbs.py
amal-ghamdi: Okay I see now, thanks for the explanation :)

amal-ghamdi (approved):
LGTM! thank you for addressing my questions.

tests/zexperimental/test_mcmc.py
```diff
@@ -788,3 +788,38 @@ def test_find_valid_samplers_implicit_prior():
...
+
+    # Joint distribution
+    joint = cuqi.distribution.JointDistribution(x, y, d, s)
+
+    # Posterior
+    posterior = joint(y=y_data)
+
+    # Sampling strategy
+    sampling_strategy = {
+        "d" : cuqi.experimental.mcmc.MH(initial_point=3),
```
chaozg, line 814: I just want to add a tiny comment that is not very relevant to this PR: initial point will be set to a `np.array` type if not specified, through [_get_default_initial_point](https://github.com/CUQI-DTU/CUQIpy/blob/4a07108f2bb3505dfa8b5db2a8b7ba1a1fde766d/cuqi/experimental/mcmc/_sampler.py#L389C1-L391C28); if specified, it will be set to whatever the user gives, a scalar in this case. I wonder if we should add something to check if the user specified initial point is compatible and possibly convert it to a uniform format, like `np.array`. 

A somehow related scenario, with the current test case, if we set the initial point of `x` to be a list, `list(0.5*np.ones(10))`, it will lead to an error:
```bash
--> 206 self.current_samples[par_name] = sampler.current_point.reshape(-1)

AttributeError: 'list' object has no attribute 'reshape'
```

chaozg (approved):
Thanks @nabriis for the nice PR and setting initial point for `HybridGibbs` is really an important feature. I only have some off-topic comment on handling initial point in general.

tests/zexperimental/test_mcmc.py
nabriis: Yes this is a good point. I suggest we add an issue to handle setting initial point more robustly.

nabriis: #487

PR #492, opened 2024-08-28
enable FD gradient in NUTS

amal-ghamdi:


amal-ghamdi:
replaced by #624 and #623

PR #497, opened 2024-08-28
Fix plot of 1D geometries

### LINKED ISSUE #322: Bug in plot() of cuqi.samples._samples.Samples
There seems a bug with the `plot` of `cuqi.samples._samples.Samples` when the samples' dimension is 1. 

```python
import cuqi

my_gaussian = cuqi.distribution.Gaussian(mean=1, cov=1)

samples = my_gaussian.sample(10)

samples.plot()
```
The above code displays the following msg and plot:
```bash
[<matplotlib.lines.Line2D at 0x1770b4b60>,
 <matplotlib.lines.Line2D at 0x1770b4b90>,
 <matplotlib.lines.Line2D at 0x1770b4c80>,
 <matplotlib.lines.Line2D at 0x17707bb60>,
 <matplotlib.lines.Line2D at 0x17707bb90>]
```

Image

chaozg:
fixed #322 

After this PR:

output

### DIFF
--- a/tests/test_samples.py
+++ b/tests/test_samples.py
@@ -54,6 +54,22 @@ def test_samples_plot(geom, to_funvals, plot_par):
     with pytest.raises(ValueError,match=r"Cannot pass is_par as a plotting argument"):
         s.plot(is_par=not to_funvals,plot_par=plot_par)
 
+@pytest.mark.parametrize("geom",[
+						(cuqi.geometry.Discrete(1)),
+						(cuqi.geometry.Continuous1D(1))
+						])
+@pytest.mark.parametrize("kwargs",[
+                        ({}),
+                        ({"marker":"o"}),
+                        ])
+def test_samples_plot(geom, kwargs, plot_par):
+    # Make basic distribution and sample
+    my_gaussian = cuqi.distribution.Gaussian(mean=1, cov=1)
+    samples = my_gaussian.sample(10)
+    samples.geometry = geom
+    fig1 = samples.plot(**kwargs)
+    fig2 = samples.plot(0, **kwargs)
+
 @pytest.mark.parametrize("to_funvals, plot_par", [(True,False),(True,True),(False,False),(False,True)])
 @pytest.mark.parametrize("plot_method, plot_par_supported, arviz",
                          [(Samples.plot_autocorrelation, False, True),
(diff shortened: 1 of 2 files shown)


amal-ghamdi (commented):
Thank you @chaozg for this helpful update. One thing to consider here is that `*args, **kwargs` are matplotlib keyword arguments. In principle, currently, in CUQIpy, one can do (this is based on your example code) and get the desired outcome.

```
import cuqi

my_gaussian = cuqi.distribution.Gaussian(mean=1, cov=1)

samples = my_gaussian.sample(10)

samples.plot(marker='o')
```
My suggestion, in the case of self.par_dim == 1, you can check if the marker kwargs argument is passed by the user or not (something like ` "marker" in kwargs.keys()`), if it is passed, it takes precedent. 

Another comment, I think now plot is called twice in case `self.par_dim == 1`.

I also suggest you add a test for the case when the user add `marker` as `kwargs` and for the case they do not.

amal-ghamdi:
> fixed #322
> 
> After this PR: output

It seems for me the figure is for par_dim=5 but your fix is only related to par_dim == 1?

chaozg:
Thank you @amal-ghamdi , I've taken your suggestion and also added a test case

tests/test_samples.py
```diff
@@ -54,6 +54,22 @@ def test_samples_plot(geom, to_funvals, plot_par):
     with pytest.raises(ValueError,match=r"Cannot pass is_par as a plotting argument"):
         s.plot(is_par=not to_funvals,plot_par=plot_par)
 
+@pytest.mark.parametrize("geom",[
+						(cuqi.geometry.Discrete(1)),
+						(cuqi.geometry.Continuous1D(1))
+						])
+@pytest.mark.parametrize("kwargs",[
+                        ({}),
+                        ({"marker":"o"}),
```
amal-ghamdi, line 63: Just a small suggestion, could use something other than the default option here, for example 
```suggestion
                        ({"marker":"*"}),
```

amal-ghamdi (approved):
Thank you @chaozg, Looks great to me. I only added a small suggestion, and I am approving now.

chaozg:
Thank you @amal-ghamdi . I'm now requesting @jakobsj 's review.

jakobsj (approved):
Nice fix and also great to keep compatible with matplotlib

PR #530, opened 2024-09-25
Checkpointing nuts

### LINKED ISSUE #404: test NUTS checkpointing/state/history
Description:
--------------
NUTS does not pass history, state and checkpointing tests. We want to fix this.

DoD:
------
- [x] Checkpointing, state, and history tests pass successfully for NUTS
- [x] Agreement on what constitutes state for NUTS is made (implied by the review and based on some discussions)
- [x] Any remaining issues (such as what is pointed out in the original description below) is delegated to other issues (comment by @amal-ghamdi: issue #537 is added and the test below is redundant I believe since we checked the checkpointing)
- [x] PR reviewed and merged

Original issue:
--------------

- Include testing case:
```
warmup(Ns)
warmup(Ns)
```
is equivalent to 
```
warmup(2*Ns)
```

Furthermore, consider getting rid of the _pre_warmup and _pre_sample methods. See Discussion in #408

amal-ghamdi:
closes #404

### DIFF
--- a/cuqi/experimental/mcmc/_hmc.py
+++ b/cuqi/experimental/mcmc/_hmc.py
@@ -98,7 +98,7 @@ class NUTS(Sampler):
                                                     'epsilon_list',
                                                     'epsilon_bar_list'})
 
-    def __init__(self, target=None, initial_point=None, max_depth=15,
+    def __init__(self, target=None, initial_point=None, max_depth=None,
                  step_size=None, opt_acc_rate=0.6, **kwargs):
         super().__init__(target, initial_point=initial_point, **kwargs)
 
@@ -158,6 +158,8 @@ class NUTS(Sampler):
 
     @max_depth.setter
     def max_depth(self, value):
+        if value is None:
+            value = 15 # default value
         if not isinstance(value, int):
             raise TypeError('max_depth must be an integer.')
         if value < 0:
--- a/cuqi/experimental/mcmc/_hmc.py
+++ b/cuqi/experimental/mcmc/_hmc.py
@@ -88,8 +88,7 @@ class NUTS(Sampler):
     """
 
     _STATE_KEYS = Sampler._STATE_KEYS.union({'_epsilon', '_epsilon_bar',
-                                                '_H_bar', '_mu',
-                                                '_alpha', '_n_alpha',
+                                                '_H_bar',
                                                 'current_target_logd',
                                                 'current_target_grad',
                                                 'max_depth'})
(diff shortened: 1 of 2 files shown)


amal-ghamdi:
@nabriis This fix enables adding NUTS in the checkpointing tests. It does not address the referred issue fully, but I think we can change the scope of the issue to "Fix NUTS checkpointing tests"

I appreciate it if you can take a look

cuqi/experimental/mcmc/_hmc.py
```diff
@@ -89,13 +89,16 @@ class NUTS(Sampler):
 
     _STATE_KEYS = Sampler._STATE_KEYS.union({'_epsilon', '_epsilon_bar',
                                                 '_H_bar', '_mu',
-                                                '_alpha', '_n_alpha'})
+                                                '_alpha', '_n_alpha',
+                                                'current_target_logd',
+                                                'current_target_grad',
+                                                'max_depth'})
```
nabriis, line 95: I feel like max_depth is not a state? I would be similar to max_it in linearRTO

nabriis (changes requested):
Looks good @amal-ghamdi. Had 1 small comment, and then I wanted to ask if you had looked at
state_history_targets?

Do you want to add NUTS to that in this PR? Its related to checkpointing.

cuqi/experimental/mcmc/_hmc.py
amal-ghamdi: Good point.

The checkpointing test below fails if I removed max depth. In checkpoint targets list below I added NUTS with specific tree depth. , The `sampler_fresh`, loses this information if max_depth is not part of the state. This is possibly up to us if we want such information to be part of the state or not. Do we want the sampler that loads the checkpoint to continue exactly as the original sampler?

```python
# List of all samplers from cuqi.experimental.mcmc that should be tested for checkpointing + their parameters
checkpoint_targets = [
    cuqi.experimental.mcmc.ULA(cuqi.testproblem.Deconvolution1D().posterior, scale=0.0001),
    cuqi.experimental.mcmc.MALA(cuqi.testproblem.Deconvolution1D().posterior, scale=0.0001),
    cuqi.experimental.mcmc.LinearRTO(cuqi.testproblem.Deconvolution1D().posterior),
    cuqi.experimental.mcmc.UGLA(create_lmrf_prior_target(dim=16)),
    cuqi.experimental.mcmc.Direct(cuqi.distribution.Gaussian(np.zeros(10), 1)),
    cuqi.experimental.mcmc.Conjugate(create_conjugate_target("Gaussian-Gamma")),
    cuqi.experimental.mcmc.ConjugateApprox(create_conjugate_target("LMRF-Gamma")),
    cuqi.experimental.mcmc.NUTS(cuqi.testproblem.Deconvolution1D(dim=10).posterior, max_depth=4)
]
    
# List of samplers from cuqi.experimental.mcmc that should be skipped for checkpoint testing


@pytest.mark.parametrize("sampler", checkpoint_targets)
def test_checkpointing(sampler: cuqi.experimental.mcmc.Sampler):
    """ Check that the checkpointing functionality works. Tested with save_checkpoint(filename) and load_checkpoint(filename).
    This also implicitly tests the get_state(), set_state(), get_history(), and set_history() as well as the reset() methods.
    
    """

    # Run sampler with some samples
    sampler.warmup(50).sample(50)

    # Save checkpoint
    sampler.save_checkpoint('checkpoint.pickle')

    # Reset the sampler, e.g. remove all samples but keep the state
    # Calling sampler.reset() would fail since it resets state.
    # Instead, we want to call a method like reset_history().
    # Currently we do this:
    sampler._samples = []

    # Do some more samples from pre-defined rng state
    np.random.seed(0)
    samples1 = sampler.warmup(50).sample(50).get_samples().samples

    #TODO Consider changing now that target=None is allowed.
    # Now load the checkpoint on completely fresh sampler not even with target
    sampler_fresh = sampler.__class__(sampler.target) # In principle init with no arguments. Now still with target
    sampler_fresh.load_checkpoint('checkpoint.pickle')

    # Do some more samples from pre-defined rng state
    np.random.seed(0)
    samples2 = sampler_fresh.warmup(50).sample(50).get_samples().samples

    # Check that the samples are the same
    assert np.allclose(samples1, samples2), f"Samples1: {samples1}\nSamples2: {samples2}"

```

amal-ghamdi:
> Looks good @amal-ghamdi. Had 1 small comment, and then I wanted to ask if you had looked at state_history_targets?
> 
> Do you want to add NUTS to that in this PR? Its related to checkpointing.

Good point, I'll change the issue scope to take care of both history and state.

cuqi/experimental/mcmc/_hmc.py
nabriis: We discussed and agree to to max_depth in state for now

nabriis (approved):
Nice! LGTM!

amal-ghamdi:
Thank you @nabriis, I added NUTS to the list of samplers tested for the history. I also made some updates to the NUTS state for simplification and removing state variables that are not needed. I appreciate it if you can take a look

chaozg (approved):
LGTM

nabriis (approved):

PR #533, opened 2024-09-26
Display acc rate on progress bar

### LINKED ISSUE #374: Use Progress bar to print sampling progress and possibly other useful sampling information
Use Progress bar to print sampling progress and possibly provide other useful sampling information (e.g. display dynamic runtime convergence diagnostics)

### LINKED ISSUE #532: None in _acc of NUTS
**Description**

The `_acc` attribute of NUTS has some extra `None` values. Maybe @amal-ghamdi you know the reason behind it?

**Example to reproduce**

```python
import cuqi
import numpy as np

gaussian = cuqi.distribution.Gaussian(np.array([0.0]), np.array([1.0]))
sampler = cuqi.experimental.mcmc.NUTS(gaussian, initial_point=np.array([0.1]))
sampler.sample(10)
sampler._acc
```

`sampler._acc` has some `None` values as seen here:
```bash
[None, 0, None, 0, None, 0, None, 0, None, 1, None, 1, None, 0, None, 1, None, 0, None, 0, None]
```

**Note**: we're drawing 10 samples, but there are 21 elements in `_acc`, including 10 extra `None`.

chaozg:
fixed #374 .

(edit be @amal-ghamdi , also closes #532 )

## Discription
So far, I only added `acc rate` information to the right end of the progress bar. 

With the following MWE,
```python
import cuqi
import numpy as np

uniform = cuqi.distribution.Uniform(np.array([0.0]), np.array([1.0]))
sampler = cuqi.experimental.mcmc.MALA(uniform, initial_point=np.array([0.1]))
sampler.warmup(10000)
sampler.sample(10000)
samples.plot_trace()
```
the progress bar will look like:
```bash
Warmup: 100%|██████████| 10000/10000 [00:08<00:00, 1144.12it/s, acc rate: 35.98%]
Sample: 100%|██████████| 10000/10000 [00:13<00:00, 722.34it/s, acc rate: 36.17%]
```
Note the `acc rate: XX.XX%` at the end.

## Questions
- Do we need/want to add any other information?
- Do we want to display the acc rate of
  - [ ] the (up-to-date) whole chain (which might be samples from multiple sample/warmups) or
  - [x] the samples from the current sample/warmup?

## Note
It seems NUTS stores some extra `None` in `_acc` and I have created an issue for it: #532 : addressed by @amal-ghamdi

### DIFF
--- a/cuqi/experimental/mcmc/_hmc.py
+++ b/cuqi/experimental/mcmc/_hmc.py
@@ -258,7 +258,7 @@ class NUTS(Sampler):
                 self.current_point = point_prime
                 self.current_target_logd = logd_prime
                 self.current_target_grad = np.copy(grad_prime)
-                acc = acc or 1
+                acc = 1
 
 
             # update number of particles, tree level, and stopping criterion
--- a/cuqi/experimental/mcmc/_sampler.py
+++ b/cuqi/experimental/mcmc/_sampler.py
@@ -231,9 +231,7 @@ class Sampler(ABC):
             self._samples.append(self.current_point)
 
             # display acc rate at progress bar
-            # workaround for None values in acc of NUTS
-            acc_filtered = list(filter(lambda x: x is not None, self._acc))
-            pbar.set_postfix_str(f"acc rate: {np.mean(acc_filtered):.2%}")
+            pbar.set_postfix_str(f"acc rate: {np.mean(self._acc):.2%}")
 
             # Add sample to batch
             if batch_size > 0:
@@ -281,9 +279,7 @@ class Sampler(ABC):
             self._samples.append(self.current_point)
 
             # display acc rate at progress bar
-            # workaround for None values in acc of NUTS
-            acc_filtered = list(filter(lambda x: x is not None, self._acc))
-            pbar.set_postfix_str(f"acc rate: {np.mean(acc_filtered):.2%}")
+            pbar.set_postfix_str(f"acc rate: {np.mean(self._acc):.2%}")
 
             # Call callback function if specified
             self._call_callback(self.current_point, len(self._samples)-1)
--- a/cuqi/experimental/mcmc/_sampler.py
+++ b/cuqi/experimental/mcmc/_sampler.py
@@ -220,7 +220,8 @@ class Sampler(ABC):
         if hasattr(self, "_pre_sample"): self._pre_sample()
 
         # Draw samples
-        for _ in tqdm( range(Ns), "Sample: "):
+        pbar = tqdm(range(Ns), "Sample: ")
+        for _ in pbar:
             
             # Perform one step of the sampler
             acc = self.step()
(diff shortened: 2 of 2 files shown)


amal-ghamdi (commented):
Thank you @chaozg! Nice display. now that acc_rate for NUTS is fixed, the workaround for NUTS can be removed I assume

cuqi/experimental/mcmc/_sampler.py
```diff
@@ -229,6 +230,9 @@ def sample(self, Ns, batch_size=0, sample_path='./CUQI_samples/') -> 'Sampler':
             self._acc.append(acc)
             self._samples.append(self.current_point)
 
+            # display acc rate at progress bar
+            pbar.set_postfix_str(f"acc rate: {np.mean(self._acc):.2%}")
```
amal-ghamdi, line 234: I have a comment here and I am not sure what is the best decision. Maybe @nabriis can also weigh in. Consider for example the code

```python
import cuqi
import numpy as np
np.random.seed(10)

uniform = cuqi.distribution.Uniform(np.array([0.0]), np.array([1.0]))
sampler = cuqi.experimental.mcmc.NUTS(uniform, initial_point=np.array([0.1]))
sampler.sample(Ns=10)
sampler.sample(Ns=10)
sampler.sample(Ns=10)
sampler.sample(Ns=10)
sampler.sample(Ns=10)
sampler.sample(Ns=2)
```

When reporting the acceptance rate, do we want to report the acceptance rate of the Ns samples or for all the samples from before? The code behavior here seems to be the latter (all the samples, including ones obtained in previous `sample` calls).  

To make it only for the Ns samples maybe we can do some updates like
```python
for sample_idx in pbar:
   # ..
   # code ..
   # .. 
    pbar.set_postfix_str(f"acc rate: {np.mean(self._acc[-1-sample_idx:]):.2%}")

```

This also applies to the warmup method

amal-ghamdi (commented):
Thank you @chaozg, I added one comment that we might consider about which samples do we report the acceptance rate for. Still not sure what is the best decision.

cuqi/experimental/mcmc/_sampler.py
chaozg: Done as suggested

amal-ghamdi (approved):
Thank you for the update, LGTM!

cuqi/experimental/mcmc/_hmc.py
```diff
@@ -260,9 +258,8 @@ def step(self):
                 self.current_point = point_prime
                 self.current_target_logd = logd_prime
                 self.current_target_grad = np.copy(grad_prime)
-                self._acc.append(1)
-            else:
-                self._acc.append(0)
+                acc = acc or 1
```
nabriis, line 261: Can you explain this change? Seems acc is initially 0, so will always be set to 1 here. Why the or?

nabriis (approved):
Hi @chaozg. LGTM. Had a question on one of the changed lines, but nothing major. You can explain when we meet next.

cuqi/experimental/mcmc/_sampler.py
nabriis: Seems reasonable to me.

cuqi/experimental/mcmc/_hmc.py
chaozg: Thanks @nabriis for spotting this typo. Indeed, this line should be `acc = 1`.

PR #534, opened 2024-09-26
Fix typo (PCN intead of pCN in experimental) in BayesianProblem

### LINKED ISSUE #528: AttributeError: module 'cuqi.experimental.mcmc' has no attribute 'pCN'
**Description**

Made custom model without derivative/gradient defined. When running UQ on Bayesian problem constructed with it, I get crash "AttributeError: module 'cuqi.experimental.mcmc' has no attribute 'pCN'". Probably a naming bug.

**Example to reproduce**

Custom model actually used annoying to put in, but would expect similar results with similar problems.
```
bp = BayesianProblem(h, m)
bp.set_data(m=synthetic_sample)
samples = bp.UQ(exact=hyp_1, experimental=True)
```

chaozg:
fixed #528

### DIFF
--- a/cuqi/problem/_problem.py
+++ b/cuqi/problem/_problem.py
@@ -614,7 +614,7 @@ class BayesianProblem(object):
 
             scale = 0.02
 
-            sampler = cuqi.experimental.mcmc.pCN(self.posterior, scale, callback=callback)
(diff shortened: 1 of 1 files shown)


nabriis (approved):
Thanks @chaozg LGTM

amal-ghamdi (approved):
Thanks for the fix @chaozg , LGTM

PR #536, opened 2024-09-30
update NUTS acc rate list update

### LINKED ISSUE #532: None in _acc of NUTS

amal-ghamdi:
This is to fix #532

### DIFF
--- a/cuqi/experimental/mcmc/_hmc.py
+++ b/cuqi/experimental/mcmc/_hmc.py
@@ -119,9 +119,6 @@ class NUTS(Sampler):
 
     def _initialize(self):
 
-        # Arrays to store acceptance rate
-        self._acc = [None] # Overwrites acc from Sampler. TODO. Check if this is necessary
-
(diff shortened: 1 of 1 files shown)


chaozg (approved):
LGTM

PR #542, opened 2024-09-30
Add (iid) truncated normal and also add grad to normal

### LINKED ISSUE #321: Add truncated normal distribution

It will be nice to have the [truncated normal distribution](https://en.wikipedia.org/wiki/Truncated_normal_distribution), which is commonly used in describing physical properties.

### LINKED ISSUE #548: Add gradient to normal
##DoD:
- [x] Add gradient() to normal distribution
- [x] Add proper unit tests
- [ ] PR reviewed by team and merged to main

chaozg:
fixed #321 to add `TruncatedNormal`
fixed #548 to add `gradient` to `Normal`
fixed #539 to add a demo on `TruncatedNormal`

## Description

- With this PR, we can draw samples from truncated iid normal distributions, i.e., truncated version of `cuqi.distribution.normal` and also use them as priors. 
- Gibbs with a gradient sampler works also fine on it.
- Plots from demo38 can be seen at #549 .

## Test with Gibbs
```python
import cuqi
import numpy as np
np.random.seed(0)
d = cuqi.distribution.Uniform(np.array([0]),np.array([1]))
A = cuqi.model.LinearModel(lambda x: x, adjoint=lambda x: x, range_geometry=1, domain_geometry=1)
x = cuqi.distribution.TruncatedNormal(A(d), std=np.array([1]), low=np.array([-1]), high=np.array([np.Inf]))

sampling_strategy = {
    "d" : cuqi.experimental.mcmc.MALA(initial_point=np.array([0.5]), scale=0.1),
    "x" : cuqi.experimental.mcmc.MALA(initial_point=np.array([1.0]), scale=0.3)
}
joint = cuqi.distribution.JointDistribution(d, x)
sampler = cuqi.experimental.mcmc.HybridGibbs(joint, sampling_strategy=sampling_strategy)
sampler.sample(10000)
samples["x"].plot_trace()
samples["d"].plot_trace()
```

(Note: here `x` is truncated from -1, just as expected)
image
image

### DIFF
--- a/cuqi/distribution/_normal.py
+++ b/cuqi/distribution/_normal.py
@@ -65,8 +65,7 @@ class Normal(Distribution):
         elif hasattr(self.mean, "gradient"): # for likelihood
             model = self.mean
             dev = val - model.forward(*args, **kwargs)
-            if isinstance(dev, numbers.Number):
-                dev = np.array([dev])
+            print(dev)
             return model.gradient(1.0/(np.array(self.std)) @ dev, *args, **kwargs)
         else:
             raise NotImplementedError("Gradient not implemented for distribution {} with location {}".format(self,self.mean))
--- a/cuqi/distribution/_normal.py
+++ b/cuqi/distribution/_normal.py
@@ -36,6 +36,9 @@ class Normal(Distribution):
     def cdf(self, x):
         return np.prod(0.5*(1 + erf((x-self.mean)/(self.std*np.sqrt(2)))))
 
+    def gradient(self, x):
+        return -(x-self.mean)/(self.std**2)
+
     def _sample(self,N=1, rng=None):
 
         """
--- a/cuqi/distribution/_normal.py
+++ b/cuqi/distribution/_normal.py
@@ -1,6 +1,7 @@
 import numpy as np
 from scipy.special import erf
 from cuqi.geometry import _get_identity_geometries
+from cuqi.utilities import force_ndarray
 from cuqi.distribution import Distribution
 
 class Normal(Distribution):
@@ -27,6 +28,24 @@ class Normal(Distribution):
         # Init specific to this distribution
         self.mean = mean
         self.std = std
+
+    @property
+    def mean(self):
+        """ Mean of the distribution """
+        return self._mean
+
+    @mean.setter
+    def mean(self, value):
+        self._mean = force_ndarray(value, flatten=True)
+
+    @property
+    def std(self):
+        """ Std of the distribution """
+        return self._std
+
+    @mean.setter
+    def std(self, value):
+        self._std = force_ndarray(value, flatten=True)
 
     def pdf(self, x):
         return np.prod(1/(self.std*np.sqrt(2*np.pi))*np.exp(-0.5*((x-self.mean)/self.std)**2))
--- a/cuqi/distribution/_normal.py
+++ b/cuqi/distribution/_normal.py
@@ -1,4 +1,5 @@
 import numpy as np
+import numbers
 from scipy.special import erf
 from cuqi.geometry import _get_identity_geometries
 from cuqi.utilities import force_ndarray
@@ -61,9 +62,14 @@ class Normal(Distribution):
             raise NotImplementedError("Gradient not implemented for distribution {} with geometry {}".format(self,self.geometry))
         if not callable(self.mean):
             return -(val-self.mean)/(self.std**2)
+        elif hasattr(self.mean, "gradient"): # for likelihood
+            model = self.mean
+            dev = val - model.forward(*args, **kwargs)
+            if isinstance(dev, numbers.Number):
+                dev = np.array([dev])
+            return model.gradient(1.0/(np.array(self.std)) @ dev, *args, **kwargs)
         else:
             raise NotImplementedError("Gradient not implemented for distribution {} with location {}".format(self,self.mean))
-
 
     def _sample(self,N=1, rng=None):
 
--- a/cuqi/distribution/_normal.py
+++ b/cuqi/distribution/_normal.py
@@ -43,7 +43,7 @@ class Normal(Distribution):
         """ Std of the distribution """
         return self._std
 
-    @mean.setter
+    @std.setter
     def std(self, value):
         self._std = force_ndarray(value, flatten=True)
 
--- a/cuqi/distribution/_normal.py
+++ b/cuqi/distribution/_normal.py
@@ -1,4 +1,5 @@
 import numpy as np
+from scipy.special import erf
 from cuqi.geometry import _get_identity_geometries
 from cuqi.distribution import Distribution
 
@@ -43,6 +44,7 @@ class Normal(Distribution):
             return -(val-self.mean)/(self.std**2)
         else:
             raise NotImplementedError("Gradient not implemented for distribution {} with location {}".format(self,self.mean))
+
 
     def _sample(self,N=1, rng=None):
 
--- a/cuqi/distribution/_normal.py
+++ b/cuqi/distribution/_normal.py
@@ -1,5 +1,5 @@
 import numpy as np
-from scipy.special import erf
+from cuqi.geometry import _get_identity_geometries
 from cuqi.distribution import Distribution
 
 class Normal(Distribution):
@@ -36,8 +36,13 @@ class Normal(Distribution):
     def cdf(self, x):
         return np.prod(0.5*(1 + erf((x-self.mean)/(self.std*np.sqrt(2)))))
 
-    def gradient(self, x):
-        return -(x-self.mean)/(self.std**2)
+    def _gradient(self, val, *args, **kwargs):
+        if not type(self.geometry) in _get_identity_geometries():
+            raise NotImplementedError("Gradient not implemented for distribution {} with geometry {}".format(self,self.geometry))
+        if not callable(self.mean):
+            return -(val-self.mean)/(self.std**2)
+        else:
+            raise NotImplementedError("Gradient not implemented for distribution {} with location {}".format(self,self.mean))
 
     def _sample(self,N=1, rng=None):
 
--- a/cuqi/distribution/_truncated_normal.py
+++ b/cuqi/distribution/_truncated_normal.py
@@ -78,6 +78,7 @@ class TruncatedNormal(Distribution):
         """
         Generates random samples from the distribution.
         """
+        # FIXME: this implementation does not honor the rng
         max_iter = 1e9 # maximum number of trials to avoid infinite loop
         samples = []
         for i in range(int(max_iter)):
--- a/cuqi/distribution/_truncated_normal.py
+++ b/cuqi/distribution/_truncated_normal.py
@@ -72,4 +72,15 @@ class TruncatedNormal(Distribution):
         """
         Generates random samples from the distribution.
         """
-        raise NotImplementedError(f"sample is not implemented for {self.__class__.__name__}.")
+        max_iter = 1e9 # maximum number of trials to avoid infinite loop
+        samples = []
+        for i in range(int(max_iter)):
+            if len(samples) == N:
+                break
+            sample = self._normal.sample()
+            if np.all(sample >= self.low) and np.all(sample <= self.high):
+                samples.append(sample)
+        # raise a error if the number of iterations exceeds max_iter
+        if i == max_iter-1:
+            raise RuntimeError("Failed to generate {} samples within {} iterations".format(N, max_iter))
+        return np.array(samples).T.reshape(-1,N)
--- a/cuqi/distribution/_truncated_normal.py
+++ b/cuqi/distribution/_truncated_normal.py
@@ -1,6 +1,7 @@
 import numpy as np
 from scipy.special import erf
 from cuqi.distribution import Distribution
+from cuqi.distribution import Normal
 
 class TruncatedNormal(Distribution):
     """
@@ -30,20 +31,23 @@ class TruncatedNormal(Distribution):
         self.low = low
         self.high = high
 
+        # Init underlying normal distribution
+        self._normal = Normal(self.mean, self.std)
+
     def logpdf(self, x):
         # the unnormalized logpdf
         # check if x falls in the range between np.array a and b
         if np.any(x < self.low) or np.any(x > self.high):
             return -np.Inf
         else:
-             return np.sum(-np.log(self.std*np.sqrt(2*np.pi))-0.5*((x-self.mean)/self.std)**2)
+             return self._normal.logpdf(x)
 
     def gradient(self, x):
         # check if x falls in the range between np.array a and b
         if np.any(x < self.low) or np.any(x > self.high):
             return np.NaN*np.ones_like(x)
         else:
-            return -(x-self.mean)/(self.std**2)
+            return self._normal.gradient(x)
 
     def _sample(self,N=1, rng=None):
         """
--- a/cuqi/distribution/_truncated_normal.py
+++ b/cuqi/distribution/_truncated_normal.py
@@ -11,32 +11,36 @@ class TruncatedNormal(Distribution):
     
     Parameters
     ------------
-    mean: mean of distribution
-    std: standard deviation
-    a: lower bound of the distribution
-    b: upper bound of the distribution
+    mean : float or array_like of floats
+        mean of distribution
+    std : float or array_like of floats
+        standard deviation
+    a : float or array_like of floats
+        lower bound of the distribution
+    b : float or array_like of floats
+        upper bound of the distribution
     """
-    def __init__(self, mean=None, std=None, a=-np.Inf, b=np.Inf, is_symmetric=False, **kwargs):
+    def __init__(self, mean=None, std=None, low=-np.Inf, high=np.Inf, is_symmetric=False, **kwargs):
         # Init from abstract distribution class
         super().__init__(is_symmetric=is_symmetric, **kwargs)  
 
         # Init specific to this distribution
         self.mean = mean
         self.std = std
-        self.a = a
-        self.b = b
+        self.low = low
+        self.high = high
 
     def logpdf(self, x):
         # the unnormalized logpdf
         # check if x falls in the range between np.array a and b
-        if np.any(x < self.a) or np.any(x > self.b):
+        if np.any(x < self.low) or np.any(x > self.high):
             return -np.Inf
         else:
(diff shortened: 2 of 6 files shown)


cuqi/distribution/_truncated_normal.py
```diff
@@ -0,0 +1,48 @@
+import numpy as np
+from scipy.special import erf
+from cuqi.distribution import Distribution
+
+class TruncatedNormal(Distribution):
+    """
+    Truncated Normal probability distribution. Generates instance of cuqi.distribution.TruncatedNormal.
```
amal-ghamdi, line 7: We can possibly add a bit more description about the distribution: e.g. 
```suggestion
    Truncated Normal probability distribution. Generates instance of cuqi.distribution.TruncatedNormal. It allows the user to specify upper and lower bounds on random variables represented by a Normal distribution. This distribution is suitable for a small dimension setup (e.g. `dim`=3 or 4). Using TruncatedNormal Distribution with a larger dimension can lead to a high rejection rate when used within MCMC samplers.
```

cuqi/distribution/_truncated_normal.py
```diff
@@ -0,0 +1,48 @@
...
+    
+    The variables of this distribution are iid.
+
+    
+    Parameters
+    ------------
+    mean: mean of distribution
+    std: standard deviation
+    a: lower bound of the distribution
+    b: upper bound of the distribution
```
amal-ghamdi, line 17: I am not sure what is the best name for the upper and lower bounds, I suggest, however, making them consistent with the uniform distribution. In the uniform distribution, we have 
```
        low : float or array_like of floats
            Lower bound(s) of the uniform distribution.
        high : float or array_like of floats 
            Upper bound(s) of the uniform distribution.
```
 It could be we also use `low` and `high` here, or we change the names in both distributions to `lower` and `upper` or something, what do you think?

cuqi/distribution/_truncated_normal.py
```diff
@@ -0,0 +1,48 @@
...
+from cuqi.distribution import Distribution
+
+class TruncatedNormal(Distribution):
+    """
+    Truncated Normal probability distribution. Generates instance of cuqi.distribution.TruncatedNormal.
+    
+    The variables of this distribution are iid.
+
+    
+    Parameters
```
amal-ghamdi, line 12: I suggest adding types to the parameters, e.g. 
```
    mean: scalar or 1d-array
        Mean of the distribution
```

cuqi/distribution/_truncated_normal.py
```diff
@@ -0,0 +1,48 @@
...
+        self.a = a
+        self.b = b
+
+    def logpdf(self, x):
+        # the unnormalized logpdf
+        # check if x falls in the range between np.array a and b
+        if np.any(x < self.a) or np.any(x > self.b):
+            return -np.Inf
+        else:
+             return np.sum(-np.log(self.std*np.sqrt(2*np.pi))-0.5*((x-self.mean)/self.std)**2)
```
amal-ghamdi, line 35: One trick we did in lognormal distribution is that we sat up an underlying Gaussian and used its logpdf and gradient. This gives more robustness in general because we relay on a single implementation that is heavily tested. Another advantage I believe is that we get the chain rule for the gradient for free in this case (e.g. when lognormal is used for likelihood for example). You can consider using this trick here as well.

cuqi/distribution/_truncated_normal.py
```diff
@@ -0,0 +1,48 @@
...
+        if np.any(x < self.a) or np.any(x > self.b):
+            return np.NaN*np.ones_like(x)
+        else:
+            return -(x-self.mean)/(self.std**2)
+
+    def _sample(self,N=1, rng=None):
+        """
+        Generates random samples from the distribution.
+        """
+        raise NotImplementedError(f"sample is not implemented for {self.__class__.__name__}.")
```
amal-ghamdi, line 48: This is a general comment, it could be nice to add some tests like the ones we have in [test_distribution.py](https://github.com/CUQI-DTU/CUQIpy/blob/main/tests/test_distribution.py). This will make sure that if someone update this code in the future, the updates still pass the test. See for example the specific tests: `test_lognormal_logpdf`, `test_gradient_lognormal_as_prior`, and `test_gradient_lognormal_as_likelihood`.

amal-ghamdi (changes requested):
Thank you @chaozg for adding this very useful distribution! I added some comments, feel free to address them as you see fit.

cuqi/distribution/_truncated_normal.py
chaozg: That's an excellent suggestion and following this strategy it will become way easier to adapt a truncated version of `cuqi.distribution.Gaussian` (instead of the iid `cuqi.distribution.Normal`)

cuqi/distribution/_truncated_normal.py
```diff
@@ -0,0 +1,48 @@
+import numpy as np
+from scipy.special import erf
+from cuqi.distribution import Distribution
+
+class TruncatedNormal(Distribution):
+    """
+    Truncated Normal probability distribution. Generates instance of cuqi.distribution.TruncatedNormal. It allows the user to specify upper and lower bounds on random variables represented by a Normal distribution. This distribution is suitable for a small dimension setup (e.g. `dim`=3 or 4). Using TruncatedNormal Distribution with a larger dimension can lead to a high rejection rate when used within MCMC samplers.
```
nabriis, line 7: Could you break this line into smaller pieces?

cuqi/distribution/_truncated_normal.py
nabriis: Sounds good to me.

cuqi/distribution/_truncated_normal.py
nabriis: Agreed

cuqi/distribution/_truncated_normal.py
```diff
@@ -0,0 +1,48 @@
...
+             return np.sum(-np.log(self.std*np.sqrt(2*np.pi))-0.5*((x-self.mean)/self.std)**2)
+
+    def gradient(self, x):
+        # check if x falls in the range between np.array a and b
+        if np.any(x < self.a) or np.any(x > self.b):
+            return np.NaN*np.ones_like(x)
+        else:
+            return -(x-self.mean)/(self.std**2)
+
+    def _sample(self,N=1, rng=None):
```
nabriis, line 44:
```suggestion
    def _sample(self, N=1, rng=None):
```

cuqi/distribution/_truncated_normal.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        Generates random samples from the distribution.
```
nabriis, line 46: Perhaps for the direct sampling of the distribution we could use a simple "Rejection" sampling? If asked for "N" samples, we sample a Gaussian, say "N" times, remove out of bounds, sample another "N" times, remote out of bounds. If we have now "N" or more sample we return the first N, else we repeat until we get "N" samples. Could also be used to compare with the samplers you showed in the code.

cuqi/distribution/_truncated_normal.py
```diff
@@ -0,0 +1,48 @@
+import numpy as np
+from scipy.special import erf
+from cuqi.distribution import Distribution
+
+class TruncatedNormal(Distribution):
```
nabriis, line 5: You made this really nice showcases in the PR @chaozg. I suggest you add one or two of them as "Examples" in the docstring!

nabriis (changes requested):
Really nice @chaozg. I agree with Amals suggestions and I had a few of my own.

cuqi/distribution/_truncated_normal.py
chaozg: Done as in uniform:
```python
    mean : float or array_like of floats
        mean of distribution
    std : float or array_like of floats
        standard deviation
    low : float or array_like of floats
        lower bound of the distribution
    high : float or array_like of floats
        upper bound of the distribution
```

cuqi/distribution/_truncated_normal.py
chaozg: Added `test_TruncatedNormal_logpdf` and `test_TruncatedNormal_gradient`

cuqi/distribution/_truncated_normal.py
chaozg: Done as suggested

cuqi/distribution/_truncated_normal.py
chaozg: I just added`_sample` with a slightly different strategy: I generate samples one by one, check if it is within the bounds, until we get N samples. Do you think strategy is OK?

cuqi/distribution/_truncated_normal.py
chaozg: Added a example in the docstring as suggested

chaozg:
Thanks @amal-ghamdi and @nabriis for your review. I think I have addressed your comments so I'm requesting your further review here.

Note that the existing Normal lacks grad(), so I also add it in this PR.
The sample() lack a way to carry rng, and I'm not sure what's the best way to do this and would like to listen to your advice at stand-up : )

cuqi/distribution/_truncated_normal.py
amal-ghamdi: 👍

cuqi/distribution/_normal.py
```diff
@@ -36,6 +36,9 @@ def logpdf(self, x):
     def cdf(self, x):
         return np.prod(0.5*(1 + erf((x-self.mean)/(self.std*np.sqrt(2)))))
 
+    def gradient(self, x):
+        return -(x-self.mean)/(self.std**2)
+
```
amal-ghamdi, line 41: This is a nice bonus that we have a gradient for normal now :). Just one suggestion to add two checks:
1- the geometry is not a geometry that needs a chain rule. 
2- The distribution is used as prior, not likelihood because we do not account for the chain rule here. 

For example, the gradient implementation of [cmrf](https://github.com/CUQI-DTU/CUQIpy/blob/main/cuqi/distribution/_cmrf.py) does these two checks



```python
    def _gradient(self, val, **kwargs):
        #Avoid complicated geometries that change the gradient.
        if not type(self.geometry) in _get_identity_geometries():
            raise NotImplementedError("Gradient not implemented for distribution {} with geometry {}".format(self,self.geometry))

        if not callable(self.location): # for prior
            diff = self._diff_op._matrix @ val
            return (-2*diff/(diff**2+self.scale**2)) @ self._diff_op._matrix
        else:
            warnings.warn('Gradient not implemented for {}'.format(type(self.location)))
```

cuqi/distribution/_truncated_normal.py
amal-ghamdi: just a comment here, would it be possible just to pass the rng to `sample = self._normal.sample(rng=rng)`?

tests/test_distribution.py
```diff
@@ -28,6 +28,73 @@ def test_Normal_sample_regression(mean,var,expected):
...
+                                                np.array([-2.0, 0.0]),
+                                                np.array([2.0, 0.0])])])
+def test_TruncatedNormal_logpdf(mean,std,low,high,points):
+    x_trun = cuqi.distribution.TruncatedNormal(mean,std,low=low,high=high)
+    x = cuqi.distribution.Normal(mean,std)
+    for point in points:
+        if np.all(point >= low) and np.all(point <= high):
+            assert x_trun.logpdf(point) == approx(x.logpdf(point))
+        else:
+            assert np.isinf(x_trun.logpdf(point))
```
amal-ghamdi, line 59: a suggestion to use `np.isneginf` instead of `np.isinf` here, to check for negative infinity.

amal-ghamdi (changes requested):
Many thanks @chaozg for the updates! it is nice that the distribution is based on Normal now and that you added the gradient for the Normal distribution. I have only few comments to consider as you see fit.

cuqi/distribution/_truncated_normal.py
amal-ghamdi: 👍

cuqi/distribution/_truncated_normal.py
amal-ghamdi: 👍

cuqi/distribution/_truncated_normal.py
amal-ghamdi: 👍

tests/test_distribution.py
chaozg: Done as suggested

cuqi/distribution/_truncated_normal.py
chaozg: I think you're right @amal-ghamdi; I just updated the code as you suggested. Thanks!

amal-ghamdi (approved):
LGTM!

cuqi/distribution/_truncated_normal.py
```diff
@@ -0,0 +1,89 @@
...
+        super().__init__(is_symmetric=is_symmetric, **kwargs)  
+
+        # Init specific to this distribution
+        self.mean = mean
+        self.std = std
+        self.low = low
+        self.high = high
+
+        # Init underlying normal distribution
+        self._normal = Normal(self.mean, self.std, is_symmetric=is_symmetric, **kwargs)
```
nabriis, line 50: I think this will create issues with Gibbs sampling. We need to have getters and setters for mean and std, and a setter for _normal.

cuqi/distribution/_truncated_normal.py
```diff
@@ -0,0 +1,89 @@
...
+        if np.any(x < self.low) or np.any(x > self.high):
+            return np.NaN*np.ones_like(x)
+        else:
+            return self._normal.gradient(x)
+
+    def _sample(self, N=1, rng=None):
+        """
+        Generates random samples from the distribution.
+        """
+        # FIXME: this implementation does not honor the rng
```
nabriis, line 77:
```suggestion

```

nabriis (changes requested):

cuqi/distribution/_truncated_normal.py
chaozg: Done as suggested

chaozg: Just added a MWE on this and corresponding plots at the bottom of my main comment

chaozg:
With @amal-ghamdi 's help today, `gradient` of TruncatedNormal finally works fine as a likelihood 😄 . And now I'm requesting further review from @nabriis .

chaozg:
Hi @amal-ghamdi since I just added a demo38 on the use of `TruncatedNormal`, I will request your further review again.

cuqi/distribution/_normal.py
```diff
@@ -42,10 +62,15 @@ def _gradient(self, val, *args, **kwargs):
             raise NotImplementedError("Gradient not implemented for distribution {} with geometry {}".format(self,self.geometry))
         if not callable(self.mean):
             return -(val-self.mean)/(self.std**2)
+        elif hasattr(self.mean, "gradient"): # for likelihood
+            model = self.mean
+            dev = val - model.forward(*args, **kwargs)
+            if isinstance(dev, numbers.Number):
+                dev = np.array([dev])
+            return model.gradient(1.0/(np.array(self.std)) @ dev, *args, **kwargs)
```
amal-ghamdi, line 70: I wonder if  "np.array" is still needed now that you have the std setter?

cuqi/distribution/_truncated_normal.py
```diff
@@ -47,7 +48,47 @@ def __init__(self, mean=None, std=None, low=-np.Inf, high=np.Inf, is_symmetric=F
...
+    def std(self):
+        """ Std of the distribution """
+        return self._std
+
+    @std.setter
+    def std(self, value):
+        self._std = force_ndarray(value, flatten=True)
+        if hasattr(self, '_normal'):
+            self._normal.std = self._std
+
```
amal-ghamdi, line 74: I wonder if we can simplify the logic a bit so we do not need to store local `_mean` and `_std`. I mean if we can set the initializer, and mean and std setters and getters to be something like this:

```python
    def __init__(self, mean=None, std=None, low=-np.Inf, high=np.Inf, is_symmetric=False, **kwargs):
        # Init from abstract distribution class
        super().__init__(is_symmetric=is_symmetric, **kwargs) 
        # Init specific to this distribution
        self.low = low
        self.high = high
        
        self._normal = Normal(mean, std, is_symmetric=True, **kwargs)

    @property
    def mean(self):
        """ Mean of the distribution """
        return self._normal.mean

    @mean.setter
    def mean(self, value):
        self._normal.mean = force_ndarray(value, flatten=True)

    @property
    def std(self):
        """ Std of the distribution """
        return self._normal.std

    @std.setter
    def std(self, value):
        self._normal.std = force_ndarray(value, flatten=True)

```

demos/demo38_truncated_normal.py
```diff
@@ -0,0 +1,77 @@
...
+sampler.sample(10000)
+samples = sampler.get_samples()
+
+plt.figure()
+samples.plot_pair()
+plt.xlim(-5, 5)
+plt.ylim(-5, 5)
+plt.gca().set_aspect('equal')
+plt.title("Samples")
+# %% Demo 2
```
amal-ghamdi, line 32:
```suggestion

# %% Demo 2
```

amal-ghamdi (commented):
Hi @chaozg , many thanks for the update and the demo! 

I made a few small comments feel free to address them as you see fit.

cuqi/distribution/_normal.py
chaozg: Yes, `dev` is already of `np.array` type. So I removed the line 68 and 69 as suggested.

cuqi/distribution/_truncated_normal.py
chaozg: This simplified code seems to cause errors in initializing the Gibbs sampler, so after talking with @amal-ghamdi, we decided to keep it as is.

amal-ghamdi: Maybe we can then just keep it as it is, unless @nabriis has other suggestions

nabriis: No I think we cannot do it much simpler

nabriis (approved):
Thanks @chaozg! Nice work. LGTM

chaozg:
Thanks, @amal-ghamdi and @nabriis for your review! I'm merging to the main now

PR #543, opened 2024-10-01
Fix nuts with gibbs

### LINKED ISSUE #512: Make NUTS fully stateful in HybridGibbs
do we lose NUTS step size here (epsilon, etc)? .. if so maybe we can print a warning

I also wonder if setting sampler.step_size = sampler.epsilon would be better than doing nothing, for NUTS. But it could be for another issue. What I mean is something like:

```python
if isinstance(sampler, NUTS): 
                sampler.initial_point = sampler.current_point
                step_size = sampler.step_size
                
.
.
if not isinstance(sampler, NUTS): # Again, special case for NUTS.
                sampler.set_state(sampler_state)
                sampler.set_history(sampler_history)           
else:
                sampler.step_size = step_size

_Originally posted by @amal-ghamdi in https://github.com/CUQI-DTU/CUQIpy/pull/503#discussion_r1750386017_

### LINKED ISSUE #524: Warning in NUTS sampler for _epsilon_bar
**Description**
The line of code, below, in NUTS is repeated twice and it generates a warning, we want to fix the line to get rid of the warning.
```python
if self._epsilon_bar == "unset": # Initial value of epsilon_bar for sampling
```
The warning:
```
/Users/amal/Documents/research_code/CUQI-DTU/CUQIpy/cuqi/experimental/mcmc/_hmc.py:315: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  if self._epsilon_bar == "unset": # Initial value of epsilon_bar for sampling
```

**Example to reproduce**

See FENIcs notebook in CUQI book.
https://github.com/CUQI-DTU/CUQI-Book/pull/74/files#diff-65b82e897cf4d25e53f93d35bcc5f582f2a6ebc08a631af478f5546e7e591fd5

**Related online discussion of this warning**
https://stackoverflow.com/questions/40659212/futurewarning-elementwise-comparison-failed-returning-scalar-but-in-the-futur

DoD
-----
- [x] Fix to get rid of the warning is implemented
- [ ] Fix is reviewed and merged
- [ ] Issue created for CUQI-Book to rerun the notebooks that display this warning

### LINKED ISSUE #537: Consider getting rid of the _pre_warmup and _pre_sample methods in NUTS
Description:
--------------
consider getting rid of the _pre_warmup and _pre_sample methods in NUTS. See the discussion in https://github.com/CUQI-DTU/CUQIpy/pull/408. This will also lead to updating the new Sampler class and HybridGibbs to remove special treatment for samplers that have pre_sample and pre_warmup

DoD:
------
- [x] investigate whether this (removing the two methods) is feasible
- [x] If removing them is feasible, remove _pre_warmup and _pre_sample from all classes mentioned above
- [x] Add regression test for NUTS within Gibbs to make sure as of this update the results are consistent with the `main` branch
- [ ] The PR is reviewed and merged

amal-ghamdi:
closes #512 
closes #537 
closes #524 

- [ ] address all issues above
- [ ] add some demo or test (`demo37_experimental_mcmc_module_Gibbs.ipynb` for example)

Results from running NUTS (before and after update) `demo37_experimental_mcmc_module_Gibbs.ipynb` with changing max depth to 15

Before (time ~ 1 min)
oldNUTS

After (time ~ 20 min)
newNUTS2

### DIFF
--- a/cuqi/experimental/mcmc/_gibbs.py
+++ b/cuqi/experimental/mcmc/_gibbs.py
@@ -1,7 +1,6 @@
 from cuqi.distribution import JointDistribution
 from cuqi.experimental.mcmc import Sampler
 from cuqi.samples import Samples, JointSamples
-from cuqi.experimental.mcmc import NUTS
 from typing import Dict
 import numpy as np
 import warnings
@@ -35,12 +34,6 @@ class HybridGibbs:
     sampler converge to the joint distribution. See e.g.
     Gelman et al. "Bayesian Data Analysis" (2014), Third Edition
     for more details.
-
-    In each Gibbs step, the corresponding sampler has the initial_point 
-    and initial_scale (if applicable) set to the value of the previous step
-    and the sampler is reinitialized. This means that the sampling is not 
-    fully stateful at this point. This means samplers like NUTS will lose
-    their internal state between Gibbs steps.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


amal-ghamdi:
#512 might be addressed separately later. Now I am closing this and I opened #545 that closes #537 and #524

PR #545, opened 2024-10-06
Remove pre_sample, pre_warmup from NUTS and fix warning 

### LINKED ISSUE #524: Warning in NUTS sampler for _epsilon_bar

### LINKED ISSUE #537: Consider getting rid of the _pre_warmup and _pre_sample methods in NUTS

amal-ghamdi:
closes #524 
closes #537 

Verification Code 1:
--------------------
Code verifies this change still gives same results as old NUTS (also this is verified by regression tests)
```python
from git import Repo
repo = Repo('./CUQIpy')
branch = repo.active_branch
# Sample with NUTS + time it
import time
import numpy as np
from cuqi.testproblem import Deconvolution1D
from cuqi.distribution import Gaussian, Gamma, JointDistribution, GMRF
from cuqi.experimental.mcmc import NUTS

# Inverse problem
np.random.seed(0) 
testproblem_obj = Deconvolution1D(dim=128, phantom='sinc', noise_std=0.001)
posterior = testproblem_obj.posterior

# NUTS
Ns=15
Nb=15
print(branch.name)
print('NUTS')
start = time.time()
np.random.seed(0)
NUTS_sampler = NUTS(posterior, step_size=None)
NUTS_sampler.warmup(Nb, tune_freq=1/Nb).sample(Ns)

end = time.time()
print('Time:', end - start)
print('samples norm:', np.linalg.norm(NUTS_sampler.get_samples().samples))
plt.plot([np.linalg.norm(NUTS_sampler.get_samples().samples[:,i]) for i in range(Ns+Nb)])
plt.ylim([0, 14])
# add branch name in title
plt.title(branch.name)
plt.figure()
plt.plot(NUTS_sampler.num_tree_node_list)
plt.title(branch.name)

```
Results main branch:

nuts_main_chain
nuts_main_tree

Results this branch:
nuts_branch_chain
nuts_branch_tree

Verification Code 2 (added as a test in the PR, for smaller sample size):
--------------------

```python
import cuqi
import numpy as np
from cuqi.distribution import Gamma, Gaussian, GMRF, JointDistribution, LMRF
from cuqi.experimental.mcmc import NUTS, HybridGibbs, Conjugate, LinearRTO, ConjugateApprox, UGLA
from cuqi.testproblem import Deconvolution1D
from git import Repo
import time
repo = Repo('../')
branch = repo.active_branch

# Forward problem
np.random.seed(0)
A, y_data, info = Deconvolution1D(dim=128, phantom='sinc', noise_std=0.001).get_components()

# Bayesian Inverse Problem
s = Gamma(1, 1e-4)
x = GMRF(np.zeros(A.domain_dim), 50)
y = Gaussian(A@x, lambda s: 1/s)

# Posterior
target = JointDistribution(y, x, s)(y=y_data)

Nb=40
sampling_strategy = {
    "x" : NUTS(max_depth=7),
    "s" : Conjugate()
}

# Here we do 10 internal steps with NUTS for each Gibbs step
num_sampling_steps = {
    "x" : 1,
    "s" : 1
}

sampler = HybridGibbs(target, sampling_strategy, num_sampling_steps)
# start time
start_time = time.time()
sampler.warmup(Nb)
sampler.sample(40)
# end time
end_time = time.time()
print(end_time - start_time)


samples["x"].plot_ci(exact=info.exactSolution)
plt.title(branch.name +" time: "+str(end_time - start_time))
```

results main branch

gibbs_main

results this branch

gibbs_branch


tests/zexperimental/test_mcmc.py
```diff
@@ -907,6 +907,66 @@ def test_HybridGibbs_updates_state_only_after_accepting_sample():
...
+    samples = sampler.get_samples()
+
+    # Read samples from reference
+    file = copy_reference("data/s_x_NUTS_within_HybridGibbs.npz")
+    reference = np.load(file)
+    reference_s = reference["s"]
+    reference_x = reference["x"]
+
+    # Compare samples
+    assert np.allclose(samples["s"].samples, reference_s, rtol=1e-3)
```
amal-ghamdi, line 965: note: on my machine, I did not need to set rtol to anything, default value worked. I needed to set it to 1e-3 to pass github actions

cuqi/experimental/mcmc/_gibbs.py
```diff
@@ -136,13 +136,7 @@ def _initialize(self):
         self._set_targets()
 
         # Initialize the samplers
-        self._initialize_samplers() 
-
-        # Run over pre-sample methods for samplers that have it
-        # TODO. Some samplers (NUTS) seem to require to run _pre_warmup before _pre_sample
-        # This is not ideal and should be fixed in the future
-        for sampler in self.samplers.values():
-            self._pre_warmup_and_pre_sample_sampler(sampler)
```
nabriis, line 145: 🎉

cuqi/experimental/mcmc/_hmc.py
```diff
@@ -212,9 +216,9 @@ def step(self):
         self._num_tree_node = 0
 
         # copy current point, logd, and grad in local variables
-        point_k = self.current_point.copy() # initial position (parameters)
+        point_k = deepcopy(self.current_point) # initial position (parameters)
```
nabriis, line 219: Are you sure we need this deep copy? One may try a more efficient operation like the "slicing trick" self.current_point[:]. We can wrap this in a method for NUTS like "self._copy_array()" or something.

it would be good to have a bit of explanation as to why it is needed.

tests/zexperimental/test_mcmc.py
nabriis: Only fair. NUTS has been difficult to test.

nabriis (changes requested):
Hi @amal-ghamdi. Really nice! I only had a comment related to the deepcopy. Perhaps we can find a more efficient method.

cuqi/experimental/mcmc/_hmc.py
amal-ghamdi: Thank you @nabriis , very good point. I think deep copy made a bit of an overhead, maybe not significant and also not significant relative to computation time. For example the Gibbs example used to run in 19.6 Sec but after the modification with deep copy it runs in 20.7 Sec.

To address this, I revert to using numpy.copy which I read is more efficient than deep copy. I handelded cases where the amount could be scalar or array with if else.

The reason I used deepcopy at the first place is because some values like logd sometimes is array and sometimes is scalar.

amal-ghamdi:
> Hi @amal-ghamdi , this PR looks good and I don't have any technical comments; but I did try the two demos on my machine and the results look great. BTW, it's good to know this trick `branch = repo.active_branch` : )

Thank you @chaozg for your review!

nabriis (approved):

PR #564, opened 2024-11-08
Regularized Gaussian extension: TV

jeverink:
Adds Total Variation regularization as option to the regularized Gaussian distribution.

PR features:

- [x] Total Variation ("tv") option with RegularizedGaussian and its variants. (1D and anisotropic 2D, depending on passed geometry)
- [x] Conditioning on the strength parameter for RegularizedGaussian (for future use with Gibbs)

Regarding tests: I propose adding a small regression test of a full hierarchical model with a TV regularized model as part of the next PR, as many smaller things might change rapidly in future smalles PRs on the Regularized Gaussian Distribution.

Example results:

`
n = 128
A, y_data, info = Deconvolution1D(dim=n, phantom='square').get_components()

x  = RegularizedGaussian(0.5*np.ones(n), 0.1, regularization = "tv", strength = 10)
y  = Gaussian(A@x, 0.001)

joint = JointDistribution(x, y)
posterior = joint(y=y_data)

sampler = RegularizedLinearRTO(posterior,
                                maxit=50,
                                penalty_parameter = 10) 
sampler.sample(100, 10)

`

image

`
M = 100
A, y_data, info = Deconvolution2D(dim=M, phantom="cookie").get_components()

x  = RegularizedGMRF(0.0*np.ones(M**2), prec = lambda d:d, regularization = "tv", strength = lambda d:d, geometry = A.domain_geometry)
x = x(d = 10)
y  = Gaussian(A@x, 0.001)

joint = JointDistribution(x, y)
posterior = joint(y=y_data)

sampler = RegularizedLinearRTO(posterior,
                                maxit=25,
                                penalty_parameter = 10) 
sampler.sample(100, 10)

`

image


cuqi/experimental/mcmc/_rto.py
```diff
@@ -197,10 +204,13 @@ def __init__(self, target=None, initial_point=None, maxit=100, stepsize="automat
         self.abstol = abstol   
         self.adaptive = adaptive
         self.maxit = maxit
+        self.inner_max_it = inner_max_it
+        self.penalty_parameter = penalty_parameter
 
     def _initialize(self):
         super()._initialize()
-        self._stepsize = self._choose_stepsize()
+        if callable(self.proximal):
```
nabriis, line 212: This logic is possibly difficult to understand in the future. I suggest you add an attribute like "_inner_solver" that can be either ADMM or FISTA (can just be a string or likewise). Then use checks like that to see which inner solver is used. Define the inner solver when the target is set.

Same holds for RegularizedGausssian

cuqi/implicitprior/_regularizedGaussian.py
```diff
@@ -57,10 +64,10 @@ class RegularizedGaussian(Distribution):
                 Upper bound of box, defaults to one
 
     regularization : string or None
-        Preset regularization. Can be set to "l1". Required for use in Gibbs in future update.
-        For "l1", the following additional parameters can be passed:
+        Preset regularization. Can be set to "l1" or 'tv'. Required for use in Gibbs in future update.
+        For "l1" or "tv", the following additional parameters can be passed:
```
nabriis, line 68: Is the zero boundary conditions described anywhere for TV? else I suggest we add it in the documentation.

cuqi/implicitprior/_regularizedGaussian.py
```diff
@@ -113,15 +128,50 @@ def _parse_regularization_input_arguments(self, proximal, projector, constraint,
...
-            strength = optional_regularization_parameters["strength"]
-            self._proximal = lambda z, gamma: ProximalL1(z, gamma*strength)
+            self._strength = optional_regularization_parameters["strength"]
+            self._proximal = lambda z, gamma: ProximalL1(z, gamma*self._strength)
             self._preset = "l1"
+        elif (isinstance(regularization, str) and regularization.lower() in ["tv"]):
+            self._strength = optional_regularization_parameters["strength"]
+            if isinstance(self.geometry, (Continuous1D)):
+                self._transformation = FirstOrderFiniteDifference(self.geometry.par_dim, bc_type='zero')
+            elif isinstance(self.geometry, (Continuous2D, Image2D)):
```
nabriis, line 141: If Image2D geometry has attribute "visual_only" as true I am not sure we want to actually use a 2D TV prior. Perhaps raise an error in this case for now.

cuqi/implicitprior/_regularizedGaussian.py
```diff
@@ -206,16 +256,18 @@ def sqrtcov(self):
     def sqrtcov(self, value):
         self.gaussian.sqrtcov = value     
     
-    def get_conditioning_variables(self):
-        return self.gaussian.get_conditioning_variables()
```
nabriis, line 210: Why is this removed?

cuqi/implicitprior/_regularizedGaussian.py
```diff
@@ -206,16 +256,18 @@ def sqrtcov(self):
...
-    
     def get_mutable_variables(self):
-        return self.gaussian.get_mutable_variables()
+        add = []
```
nabriis, line 260: Cool to have strength as a mutable variable! I suggest this a bit refactoring
```python
def get_mutable_variables(self):
    mutable_vars = self.gaussian.get_mutable_variables()
    
    if self.preset in self.regularization_options():
        mutable_vars += ["strength"]
    
    return mutable_vars
```

cuqi/implicitprior/_regularizedGaussian.py
```diff
@@ -206,16 +256,18 @@ def sqrtcov(self):
...
+        if self.preset in self.regularization_options():
+            add = ["strength"]
+        return self.gaussian.get_mutable_variables() + add
     
     # Overwrite the condition method such that the underlying Gaussian is conditioned in general, except when conditioning on self.name
     # which means we convert Distribution to Likelihood or EvaluatedDensity.
     def _condition(self, *args, **kwargs):
-
+        if self.preset in self.regularization_options():
+            return super()._condition(*args, **kwargs)
```
nabriis, line 269: There seem to be some potential issues here. Namely, what if both the mean and strength were conditioning variables? Then we would only condition on the strength now. E.g.
```python
x = RegularizedGaussian(lambda x: x, 1, strength=lambda s: s, ...) 
x(strength=1, mean=2)
```
I suggest moving this check after creating "new_density" at the end, then doing the super()._condition(*args, **kwargs) on new_density.

It could also be worthwhile to add a test on this new conditioning feature.

nabriis (changes requested):
Hi @jeverink. Really nice and clean PR. I only had some minor comments on docstrings and then one main area for comments on the code for conditioning which was now updated. Let me know if you have any questions

cuqi/experimental/mcmc/_rto.py
```diff
@@ -210,8 +220,8 @@ def validate_target(self):
         super().validate_target()
         if not isinstance(self.target.prior, (cuqi.implicitprior.RegularizedGaussian, cuqi.implicitprior.RegularizedGMRF)):
             raise TypeError("Prior needs to be RegularizedGaussian or RegularizedGMRF")
-        if not callable(self.proximal):
-            raise TypeError("Proximal needs to be callable")
+        if not callable(self.proximal) and not isinstance(self.proximal, list):
+            raise TypeError("Proximal needs to be callable or a list")
```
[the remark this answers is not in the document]
jeverink: I have updated the error message

cuqi/experimental/mcmc/_rto.py
jeverink: I have now added the _inner_solver attribute.

I do not understand what you mean with 
> Same holds for RegularizedGausssian

cuqi/implicitprior/_regularizedGaussian.py
jeverink: Good catch. I have now changed the boundary condition. Weirdly the boundary condition had to be set to "neumann" for there to be no boundary conditions, instead of "none".

cuqi/implicitprior/_regularizedGaussian.py
jeverink: I have now added this condition

cuqi/implicitprior/_regularizedGaussian.py
jeverink: The first thing that `get_conditioning_variables` does is ask for `get_mutable_variables`. So as long as `get_mutable_variables` and the attributes are correct, this wrapping should not be needed anymore.

cuqi/implicitprior/_regularizedGaussian.py
jeverink: Thanks for the suggestion.

cuqi/implicitprior/_regularizedGaussian.py
jeverink: The precise workings of the conditioning is still a little bit of a mystery to me, but the issue you suggest, where there are anonymous functions for both Gaussian parameters and regularization parameters, seems to be working just fine. Also, in the full version of the code that I have, including conjugacy, this seems to work as well. 

image



I do get that this is not the best solution, however your suggestion,
if I interpret it as removing the initial check and replacing the final line with
`return new_density._condition(**kwargs)` crashes the kernel.

I have now added two tests, one with constraints and one with tv, to test the mutable and conditioning variables and verify that the conditioning does work as expected.

jeverink:
Thanks for the feedback @nabriis ,

The conditioning of the regularized Gaussian distributions is still a little bit of a mystery to me, but it all does seem to work. I hope you have a better understanding of it to know how to properly improve it.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -169,12 +169,19 @@ class RegularizedLinearRTO(LinearRTO):
         Initial point for the sampler. *Optional*.
```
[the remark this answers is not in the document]
nabriis: Great thanks!

cuqi/experimental/mcmc/_rto.py
nabriis: Thanks. Looks good. In both classes.

cuqi/implicitprior/_regularizedGaussian.py
nabriis: You are right!

cuqi/implicitprior/_regularizedGaussian.py
nabriis: Thank you. I look into it a bit more. It seems to be working as expected though

tests/test_implicit_priors.py
```diff
@@ -104,3 +104,37 @@ def test_RegularizedUnboundedUniform_is_RegularizedGaussian():
...
+    
+    assert x.get_mutable_variables() == ['mean', 'prec']
+    assert x.get_conditioning_variables() == ['a', 'b']
+
+    x = x(a=1, b=2)
+
+    assert np.allclose(x.mean, [1, 1, 1, 1])
+    assert np.allclose(x.prec, 10)
+
+def test_RegularizedGaussian_conditioning_strength():
```
amal-ghamdi, line 124: Nice!

amal-ghamdi (changes requested):
Thank you @jeverink for the great work here and expanding the features of the regularized linear RTO! I have some comments here and there for you to address as you see fit. 

One additional aspect I wanted to mention, if there is a way to simplify setting constriants/projection/ etc. Now we have: "proximal, projector, constraint, regularization" where proximal can be a list. Can everything be unified in one list for example? If this makes sense to you but is a big change, I suggest adding your thoughts on it in an issue.

cuqi/implicitprior/_regularizedGaussian.py
```diff
@@ -39,10 +41,15 @@ class RegularizedGaussian(Distribution):
...
-        min_z 0.5||x-z||_2^2+scale*g(x).
-
+    proximal : callable f(x, scale), list of tuples (callable proximal operator of f_i, linear operator L_i) or None
+        If callable:
+            Euclidean proximal operator f of the regularization function g, that is, a solver for the optimization problem
+            min_z 0.5||x-z||_2^2+scale*g(x).
+        If list of tuples (callable proximal operator of f_i, linear operator L_i):
+            Each callable proximal operator of f_i accepts two arguments (x, p) and should return the minimizer of p/2||x-z||^2 + f(x) over z for some f.
+            The corresponding regularization takes the form
+                sum_i f_i(L_i x),
```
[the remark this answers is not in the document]
jeverink: This is incorrect.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -210,8 +224,9 @@ def validate_target(self):
         super().validate_target()
         if not isinstance(self.target.prior, (cuqi.implicitprior.RegularizedGaussian, cuqi.implicitprior.RegularizedGMRF)):
             raise TypeError("Prior needs to be RegularizedGaussian or RegularizedGMRF")
-        if not callable(self.proximal):
-            raise TypeError("Proximal needs to be callable")
+        if not callable(self.proximal) and not isinstance(self.proximal, list):
+            raise TypeError("Proximal in the regularized Gaussian prior needs to be callable or a list")
```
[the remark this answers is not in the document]
jeverink: When it is a list, it is specified in the exact same way as for ADMM, as this is what is passed on to ADMM.

But I agree that this error message is superfluous, given that we check whether the prior is a RegularizedGaussian or RegularizedGMRF directly above, which constructs and/or verifies these precise conditions.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -159,6 +159,10 @@ class RegularizedLinearRTO(LinearRTO):
     Regularized Linear RTO (Randomize-Then-Optimize) sampler.
 
     Samples posterior related to the inverse problem with Gaussian likelihood and implicit Gaussian prior, and where the forward model is Linear.
+    The sampler works by repeatedly solving regularized linear least squares problems.
+    The solver for these optimization problems is chosen based on how the regularized is provided in the implicit Gaussian prior.
+    See :class:`~cuqi.implicitprior.RegularizedGaussian` for details.
```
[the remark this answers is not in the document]
jeverink: I have now replace it with a comment on the conditions under which solver is used.

cuqi/implicitprior/_regularizedGaussian.py
```diff
@@ -89,10 +97,17 @@ def _parse_regularization_input_arguments(self, proximal, projector, constraint,
...
+                if len(get_non_default_args(proximal)) != 2:
+                    raise ValueError("Proximal should take 2 arguments.")
+            if isinstance(proximal, list):
+                for (prox, op) in proximal:
+                    if len(get_non_default_args(prox)) != 2:
+                        raise ValueError("Proximal should take 2 arguments.")
+                    if op.shape[1] != self.geometry.par_dim:
+                        raise ValueError("Incorrect shape of linear operator.")
+            else:
+                raise ValueError("Proximal needs to be callable or a list. See documentation.")
```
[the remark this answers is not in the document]
jeverink: I have now created a setter for the proximal with these checks, and made sure it works with the presets.

jeverink:
Thanks for the feedback @amal-ghamdi ,
regarding the possible unification of the penalty terms,
I think that is a great idea to streamline the features, I have added the issue #584.

Also to @nabriis ,
I have now added a general issue #585 with a bit of a roadmap of the upcoming/planned features that will hopefully follow from this PR.

amal-ghamdi (approved):
Thank you @jeverink for the updates. I like how you added more details to the documentation to clarify more concepts to the user, especially in what concerns constraints/regularizers/proximals. Also thanks for creating #584 .

nabriis (approved):

nabriis:
@jeverink 🎉

PR #572, opened 2024-11-14
Add how-to on Image2D

### LINKED ISSUE #551: Can we make linear model with 2D Image geometry more intuitive
Description:
------------
As suggested by @chaozg, it seems to be a recurring issue with users to find the right setting for 2D geometry when working with linear models (in particular setting the flag "visual_only")

There is a need to clear some confusion the user might encounter when preparing their forward model: If their forward model works on 2D images, they probably need `visual_only=False` (the default) when defining the domain geometry, and they probably need `visual_only=True` when the model works on vectors.

DoD
-----
- [x] A solution is found to make this setup more intuitive to the user.
    - It is decided to add a how-to to demonstrate/differentiate the use of Image2D
- [x] In a how-to file, demonstrate when `visual_only` is used and when it is not. 
    - The how-to should be concise and the examples should be intuitive.
    - The effect of using `visual_only` should be properly demonstrated and documented.
- [ ] Guide the user to the how-to file by putting links at the following places
    - Docstring of Model
    - Docstring of Image2D
    - Docstring of CUQI array and CUQI sample
- [ ] The solution is reviewed and approved

chaozg:
fixed #551 

This PR aims to clear some confusion the user might encounter when preparing their forward model. If their forward model works on 2D images, they probably need `visual_only=False` (the default) when defining the domain geometry, and they probably need `visual_only=True` when the model works on vectors.

To deliver this information, this PR added a how-to to demonstrate two use-cases of Image2D:
- visual-only=False for image-based forward operation
- visual-only=True for vector-based forward operation

With this x_sample
image

the first case will give output:
image

while the second will give output:
image

### DIFF
--- a/demos/howtos/Image2D.py
+++ b/demos/howtos/Image2D.py
@@ -2,7 +2,7 @@
 The use of Image2D
 ==================
 
-This script demonstrates the use of :class:`~cuqi.geometry.Image2D` for domain and range geometries in CUQIpy forward models. The settings of this geometry, in particular the so-called 'visual_only' flag, depend on how the forward model input is operated on in the forward model. `visual_only` flag tells the geometry to interpret the variable as a 2D image only when visualizing the variable but keep it as a vector when being operated on in the forward model. Generally, if the forward operator expects an image input, set `visual_only` to `False`, which is the default in CUQIpy. Otherwise, if it expects a vector input, set `visual_only` to `True` in the domain geometry. In the same way, the range geometry `visual_only` setting should be consistent with your forward operator output.
+This script demonstrates the use of :class:`~cuqi.geometry.Image2D` for domain and range geometries in CUQIpy forward models. The settings of this geometry, in particular the so-called `visual_only` flag, depend on how the forward model input is operated on in the forward model. `visual_only` flag tells the geometry to interpret the variable as a 2D image only when visualizing the variable but keep it as a vector when being operated on in the forward model. Generally, if the forward operator expects an image input, set `visual_only` to `False`, which is the default in CUQIpy. Otherwise, if it expects a vector input, set `visual_only` to `True` in the domain geometry. In the same way, the range geometry `visual_only` setting should be consistent with your forward operator output.
 """
 
 # %%
--- a/demos/howtos/Image2D.py
+++ b/demos/howtos/Image2D.py
@@ -1,6 +1,6 @@
 """
 The use of Image2D
-=====================================================
+==================
 
 This script demonstrates the use of :class:`~cuqi.distribution.Image2D`. Depending on how the input variable is used in the forward model, the geometry of it can be set to be visual only or not by setting the `visual_only` attribute. Generally, if your forward operator expects a vector input, set `visual_only` to `True`. If it expects an image input, set `visual_only` to `False`, which is the default.
 """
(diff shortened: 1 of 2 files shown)


demos/howtos/Image2D.py
```diff
@@ -0,0 +1,59 @@
...
+import matplotlib.pyplot as plt
+
+# %%
+# Define dimensions
+dim_x, dim_y = 2, 4
+
+# %%
+# Example with `visual_only=False`
+# With `visual_only=False`, the underlying structure of a CUQI array or sample will be reshaped to the shape of the specified geometry, i.e., an image of size (dim_x, dim_y), before being passed to the model.
+# Here we intend to define a forward operator that downsamples the input image by a factor of 2. It's easier to define such a forward operator in terms of an image, so we set `visual_only` to `False`.
```
nabriis, line 21: A bit of a styling thing, but I feel like it may be easier to digest if we initially show the case without setting the visual only. E.g. by explaining how Image2D works in general. Then afterwards give options if the user has e.g. forward model that takes vectors

nabriis (changes requested):

demos/howtos/Image2D.py
amal-ghamdi: I think this is a good point. And I think it can be achieved with not so many changes. The howto already starts with the default behavior (visual_only=False). My suggestion here is a few changes like below could work.
```python
# Example illustrating the default behavior: `visual_only=False`
# With `visual_only=False`, the underlying structure of a CUQI array or sample will be reshaped to the
# shape of the specified geometry, i.e., an image of size (dim_x, dim_y), before being passed to the model.
# Here we intend to define a forward operator that downsamples the input image by a factor of 2. It's easier
# to define such a forward operator in terms of an image. We do not need to explicitly set `visual_only` to 
#`False` because it is the default behavior.
```

Then in defining the geometry, you can skip setting the flag for this case.

demos/howtos/Image2D.py
```diff
@@ -0,0 +1,59 @@
...
+# Plot the original sample
+plt.figure()
+x_sample.plot(cmap='Greens')
+plt.title("X image")
+
+# Apply the model and plot the result
+plt.figure()
+y_sample = model_A @ x_sample
+y_sample.plot(cmap='Greens')
+plt.title("Y image (after Model A)")
```
amal-ghamdi, line 43: I would suggest figure titles something like:
for input: "input image x"
for output: "output y=model_A(x)"

amal-ghamdi (commented):
Thank you @chaozg for this useful demo! I made some comments for your consideration.

demos/howtos/Image2D.py
chaozg: Done

chaozg:
Thanks @nabriis and @amal-ghamdi for your comments! I think I have addressed them and am now requesting your further review.

nabriis (approved):

demos/howtos/Image2D.py
```diff
@@ -0,0 +1,70 @@
+"""
+The use of Image2D
+==================
+
+This script demonstrates the use of :class:`~cuqi.geometry.Image2D` for domain and range geometries in CUQIpy forward models. The settings of this geometry, in particular the so-called 'visual_only' flag, depend on how the forward model input is operated on in the forward model. `visual_only` flag tells the geometry to interpret the variable as a 2D image only when visualizing the variable but keep it as a vector when being operated on in the forward model. Generally, if the forward operator expects an image input, set `visual_only` to `False`, which is the default in CUQIpy. Otherwise, if it expects a vector input, set `visual_only` to `True` in the domain geometry. In the same way, the range geometry `visual_only` setting should be consistent with your forward operator output.
```
amal-ghamdi, line 5:
```suggestion
This script demonstrates the use of :class:`~cuqi.geometry.Image2D` for domain and range geometries in CUQIpy forward models. The settings of this geometry, in particular the so-called `visual_only` flag, depend on how the forward model input is operated on in the forward model. `visual_only` flag tells the geometry to interpret the variable as a 2D image only when visualizing the variable but keep it as a vector when being operated on in the forward model. Generally, if the forward operator expects an image input, set `visual_only` to `False`, which is the default in CUQIpy. Otherwise, if it expects a vector input, set `visual_only` to `True` in the domain geometry. In the same way, the range geometry `visual_only` setting should be consistent with your forward operator output.
```

amal-ghamdi (approved):
Thank you @chaozg for the updates. Looks great to me. I only added one suggestion (I used the wrong quotation in my previous suggestion and this suggestion fixes it). Approved :)

demos/howtos/Image2D.py
amal-ghamdi: 👍

Issue #577, opened 2024-11-16
RV: "Basic" Random Variable class


nabriis:
Implement a "basic" random variable class with the following features:
- [x] Support for variable algebra via #576 
- [x] Support for basic stochastic elements: Knowledge of distribution
- [x] Repr showing random variable
- [x] Basic distribution related attributes like "dim, geometry"
- [x] Name attribute as with Distributions
- [x] Can be used as "drop-in" replacement of lambda function in Bayesian model
- [x] Ensure unused keywords are caught and throw and error (from discussion in https://github.com/CUQI-DTU/CUQIpy/pull/583)
- [ ] After implementation, ensure variables classes are not exposed publically to users (also from https://github.com/CUQI-DTU/CUQIpy/pull/583)
- [x] Unit testss
- [x] Updated howto

For example one should be able to do as follows:
```python
A = LinearModel(...)
s = RandomVariable(Gamma(1, 1e-4))
x = RandomVariable(Gaussian(np.zeros(128), 1))
y = RandomVariable(Gaussian(A@x, 1/s)
joint = JointDistribution(x, y)
```
For now, let us allow JointDistribution to accept a both Distribution or RandomVariable as input, then internally extract the Distribution from the RandomVariable to keep the remainder of the flow in CUQIpy the same.

Later tasks will incorporate RandomVariables more directly.

nabriis:
@jakobsj DoD

jakobsj:
@nabriis DoD approved!

PR #583, opened 2024-11-17
Add 576 variable class

### LINKED ISSUE #576: RV: Variable class with algebra (using AST)
Implement a variable class that can "record" various operations applied to it and allow this information to be used later. For example:
```python
x = Variable()
z = (x+2)**2
```
The variable `z` should know of the operations applied and it should be possible to apply them again to specific values of `x`, say `z(x=1)` should return the value 9.

- [x] Implement abstract syntax tree for a basic variable "node" class
- [x] Support core mathematical operations like (0, -, /, *)
- [x] Support mathematical operations: (abs, pow, neg, matmul)
- [x] Support operations: (getitem)
- [x] Implement repr for displaying operations applied to variable
- [x] Unit tests
- [x] Howto

nabriis:
Closes #576

### DIFF
--- a/cuqi/experimental/algebra/_ast.py
+++ b/cuqi/experimental/algebra/_ast.py
@@ -1,0 +1,294 @@
+"""
+CUQIpy specific implementation of an abstract syntax tree (AST) for algebra on variables.
+
+The AST is used to record the operations applied to variables allowing a delayed evaluation
+of said operations when needed by traversing the tree with the __call__ method.
+
+For example, the following code
+
+    x = VariableNode('x')
+    y = VariableNode('y')
+    z = 2*x + 3*y
+
+will create the following AST:
+
+z = AddNode(
+        MultiplyNode(
+            ValueNode(2),
+            VariableNode('x')
+        ),
+        MultiplyNode(
+            ValueNode(3),
+            VariableNode('y')
+        )
+    )
+
+which can be evaluated by calling the __call__ method:
+
+    z(x=1, y=2) # returns 8
+
+"""
+
+from abc import ABC, abstractmethod
+
+convert_to_node = lambda x: x if isinstance(x, Node) else ValueNode(x)
+""" Converts any non-Node object to a ValueNode object. """
+
+# ====== Base classes for the nodes ======
+
+
+class Node(ABC):
+    """Base class for all nodes in the abstract syntax tree.
+
+    Responsible for building the AST by creating nodes that represent the operations applied to variables.
+
+    Each subclass must implement the __call__ method that will evaluate the node given the input parameters.
+
+    """
+
+    @abstractmethod
+    def __call__(self, **kwargs):
+        """Evaluate node at a given parameter value. This will traverse the tree and evaluate it given the recorded operations."""
+        pass
+
+    def __add__(self, other):
+        return AddNode(self, convert_to_node(other))
+
+    def __radd__(self, other):
+        return AddNode(convert_to_node(other), self)
+
+    def __sub__(self, other):
+        return SubtractNode(self, convert_to_node(other))
+
+    def __rsub__(self, other):
+        return SubtractNode(convert_to_node(other), self)
+
+    def __mul__(self, other):
+        return MultiplyNode(self, convert_to_node(other))
+
+    def __rmul__(self, other):
+        return MultiplyNode(convert_to_node(other), self)
+
+    def __truediv__(self, other):
+        return DivideNode(self, convert_to_node(other))
+
+    def __rtruediv__(self, other):
+        return DivideNode(convert_to_node(other), self)
+
+    def __pow__(self, other):
+        return PowerNode(self, convert_to_node(other))
+
+    def __rpow__(self, other):
+        return PowerNode(convert_to_node(other), self)
+
+    def __neg__(self):
+        return NegateNode(self)
+
+    def __abs__(self):
+        return AbsNode(self)
+
+    def __getitem__(self, i):
+        return GetItemNode(self, convert_to_node(i))
+
+    def __matmul__(self, other):
+        return MatMulNode(self, convert_to_node(other))
+
+    def __rmatmul__(self, other):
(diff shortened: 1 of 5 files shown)


chaozg (approved):
Thank you @nabriis for the very interesting PR. I just want to comment that this PR reminds me a lot of the SymPy package and I'm not sure you can use SymPy as backend or draw some inspiration from there. For example, the following naive snippet  can do the conditioning on `x=1` and return a second expression after conditioning (that we can evaluate later).

```python
from sympy import *
x, y, z = symbols("x y z")
expr = x+exp(y)+log(z)
expr.subs(x, 1)
#return: y+z+1
```
Probably a bonus of using SymPy is that we would be able to do symbolic differentiation (though I'm not sure if this will be useful in the context of this PR):
```python
from sympy import *
x, y, z = symbols("x y z")
expr = x+exp(y)+log(z)
diff(expr, y)
#output: e^y
```

These are just my __naive__ thoughts and please feel free to ignore them if they are not relevant.

demos/howtos/algebra.py
```diff
@@ -0,0 +1,66 @@
...
+# algebra.
+
+print("Basic operations:")
+print(x + 1)
+print(x + y)
+print(x * y)
+print(x / y)
+
+print("\nComplex operations:")
+print(x**2 + 2*x*y + y**2)
```
amal-ghamdi, line 40: small comment about printing, I think it will be nicer if there is no spaces between the power symbol and what comes before and after

x ^ 2 vs. x^2

cuqi/experimental/algebra/_ast.py
```diff
@@ -0,0 +1,315 @@
...
+
+    Responsible for building the AST by creating nodes that represent the operations applied to variables.
+
+    Each subclass must implement the __call__ method that will evaluate the node given the input parameters.
+
+    """
+
+    @abstractmethod
+    def __call__(self, **kwargs):
+        """Evaluate node at a given parameter value. This will traverse the tree and evaluate it given the recorded operations."""
```
amal-ghamdi, line 51: is this true, below? if so it might a bit of clarity
```suggestion
        """Evaluate node at a given parameter value. This will traverse the sub-tree originated at this node and evaluate it given the recorded operations."""
```

cuqi/experimental/algebra/_ast.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+
+    def __init__(self, name):
+        self.name = name
+
+    def __call__(self, **kwargs):
+        """Retrieves the value of the variable from the passed kwargs. If no value is found, it raises a KeyError."""
+        if not self.name in kwargs:
+            raise KeyError(
+                f"Variable '{self.name}' not found in the given input parameters. Unable to evaluate the expression."
```
amal-ghamdi, line 180: do you think catching unmatched keyword arguments should generate and error? for example:

`(x+1)(x=1, y=3)`

cuqi/experimental/algebra/_ast.py
amal-ghamdi, line 1: great structure! I am wondering if we can make many of the classes here private classes if we do not intend for the user to use them explicitly

amal-ghamdi (commented):
Thank you @nabriis for this PR, very neat! I only have a few comments for your consideration.

cuqi/experimental/algebra/_ast.py
```diff
@@ -0,0 +1,315 @@
...
+    def __rmul__(self, other):
+        return MultiplyNode(convert_to_node(other), self)
+
+    def __truediv__(self, other):
+        return DivideNode(self, convert_to_node(other))
+
+    def __rtruediv__(self, other):
+        return DivideNode(convert_to_node(other), self)
+
+    def __pow__(self, other):
```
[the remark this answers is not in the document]
nabriis: At some point yes. However in native python there is not a way to implement "log" without using e.g. the "math" library. Currently this first PR focused on the base stuff supported in python without invoking a library

demos/howtos/algebra.py
```diff
@@ -0,0 +1,66 @@
...
+print((x + y)**2)
+
+print("\nProgrammatric operations:")
```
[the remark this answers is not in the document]
nabriis: Yes that was mostly why

demos/howtos/algebra.py
nabriis: Fixed

cuqi/experimental/algebra/_ast.py
nabriis: I think that would be good, although I am not sure how to handle that currently using this "localized" structure. In the random variables we can catch it, but here I am not sure how to do it.

cuqi/experimental/algebra/_ast.py
nabriis: We could yes. We may not even intend the user to use any of then. I have updated the next issue on random variables to make sure we deal with that.

cuqi/experimental/algebra/_ast.py
nabriis: I added the point to random variables. I think it is better dealt with there after we look into that task

nabriis: #577

cuqi/experimental/algebra/_ast.py
nabriis: #577

nabriis:
Thank you @chaozg and @amal-ghamdi. I believe I have answered all your questions.

As for the suggestion to use sympy. I agree this is a good idea to look into. I suggest we first implement the random variable class and a few of the other tasks, then look into using a framework like sympy to replace the AST. See https://github.com/CUQI-DTU/CUQIpy-admin/issues/22

cuqi/experimental/algebra/_ast.py
amal-ghamdi: Sounds good plan!

amal-ghamdi (approved):
Thank you for the updates. Looks great!

PR #589, opened 2024-11-25
Make the interface of restorator more robust

### LINKED ISSUE #588: Better documentation/interface on RestorationPrior
## Issue
- With the current implementation of `RestorationPrior`, we expect the user-supplied `restorator` to return a tuple of two variables, i.e., the "restored" solution and extra information.
- It's quite easy for the user (and us...) to make a mistake here as they might provide a function that only returns the "restored" solution when there is no meaningful extra information to provide. Particularly in a 2D case, we won't even notice this error as the code runs with such an incorrectly defined restorator (but gives wrong results, of course).

## DoD
- [x] Propose a solution and discuss it with team
  - Option 1: CUQIpy internally detects and handles no matter if the user-supplied restorator returns a solution or a tuple
  - Option 2: CUQIpy accepts only a function that returns a tuple of length 2 and raises error if it doesn't
  - Stand-up suggests that option 1 might add confusion to the user and we go with option 2
- [x] Investigate the use of `inspect` module to infer the return type of a user-supplied restorator
  - `inspect` seems not what we want here and we shouldn't expect our user to annotate their functions 
- [x] Check return type at run time and raise value error if the supplied restorator doesn't return a tuple of length 2.
- [x] Add unit test to check of the value error is correctly catpured
- [x] Update the docstring of `RestorationPrior` class, the `restore` method and ValueError message to guide the user.

chaozg:
fixed #588 

- add more documentation  on the required return type of the user-supplied `restorator`
- raise value error if the `restorator` doesn't return a tuple of length 2

### DIFF
--- a/cuqi/implicitprior/_restorator.py
+++ b/cuqi/implicitprior/_restorator.py
@@ -45,7 +45,7 @@ class RestorationPrior(Distribution):
     def restore(self, x, restoration_strength):
         """This function allows us to restore the input x with the user-supplied
         restorator. Extra information about the restoration operation is stored
-        in the self.info attribute.
+        in the `RestorationPrior` info attribute.
         
         Parameters
         ---------- 
--- a/cuqi/implicitprior/_restorator.py
+++ b/cuqi/implicitprior/_restorator.py
@@ -64,12 +64,11 @@ class RestorationPrior(Distribution):
         if type(restorator_return) == tuple:
             solution, self.info = restorator_return
         else:
-            raise ValueError("The restorator should return a two-element tuple" 
-                             + "with the restored parameter as the first element"
-                             + "and additional information about the restoration"
-                             + "as the second element. As to the second element,"
-                             + "it can be of any type, including `None` in case" 
-                             + "there is no information.")
+            raise ValueError("Wrong return type from the restorator. "+ 
+                             "Please ensure that the restorator returns a two-element tuple with the "+
+                             "restored solution as the first element and additional information about the "+
+                             "restoration as the second element. As to the second element, it can be of any type, "+
+                             "including `None` in case there is no particular information.")
 
         return solution
     
--- a/cuqi/implicitprior/_restorator.py
+++ b/cuqi/implicitprior/_restorator.py
@@ -16,7 +16,8 @@ class RestorationPrior(Distribution):
     ---------- 
     restorator : callable f(x, restoration_strength)
         Function f that accepts input x to be restored and returns the
-        restored version of x and information about the restoration operation.
+        restored version of x or a two-element tuple of the restored version of
+         x and extra information about the restoration operation.
             
     restorator_kwargs : dictionary
         Dictionary containing information about the restorator.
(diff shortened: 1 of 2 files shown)


chaozg:
Thank you @nabriis and @jakobsj  for your suggestion at stand-up, particularly on the possible use of the `inspect` module.
I just did some experiments on the use of it and would like to report to you that
- For `inspect` to get information of the return type of a function, the user has to annotate the `restorator` function with type hinting. For example, with the following two defined functions, we won't be able to infer the return type of `my_proximal_1 ` with `inspect`:
```python
import inspect

def my_proximal_1(x, restoration_strength):
    return np.maximum(x, 0), None

def my_proximal_2(x: np.ndarray, restoration_strength: float) -> tuple:
    return np.maximum(x, 0), None

sig_1 = inspect.signature(my_proximal_1)
print(sig_1.return_annotation)
# <class 'inspect._empty'>

sig_2 = inspect.signature(my_proximal_2)
print(sig_2.return_annotation)
# <class 'tuple'>
```
- Then I'm thinking maybe it's enough to raise a value error at run time if the user-provided function doesn't return a tuple of length 2? Of course, we should give informative instructions in the docstring of the `RestorationPrior` class, the `restore` method and also in the ValueError message.

cuqi/implicitprior/_restorator.py
```diff
@@ -41,8 +43,9 @@ def __init__(self, restorator, restorator_kwargs
         super().__init__(**kwargs)
 
     def restore(self, x, restoration_strength):
-        """This function allows us to restore the input x and returns the
-        restored version of x.
+        """This function allows us to restore the input x with the user-supplied
+        restorator. Extra information about the restoration operation is stored
+        in the self.info attribute.
```
amal-ghamdi, line 48:
```suggestion
        in the `RestorationPrior` info attribute.
```

cuqi/implicitprior/_restorator.py
```diff
@@ -54,9 +57,18 @@ def restore(self, x, restoration_strength):
...
+
+        if type(restorator_return) == tuple and len(restorator_return) == 2:
+            solution, self.info = restorator_return
+        else:
+            raise ValueError("Wrong return type from the restorator. "+ 
+                             "Please ensure that the restorator returns a two-element tuple with the "+
+                             "restored solution as the first element and additional information about the "+
+                             "restoration as the second element. As to the second element, it can be of any type, "+
+                             "including `None` in case there is no particular information.")
+
```
amal-ghamdi, line 71:
```suggestion
            raise ValueError("Unsupported return type from the user-supplied restorator function. "+ 
                             "Please ensure that the restorator function returns a two-element tuple with the "+
                             "restored solution as the first element and additional information about the "+
                             "restoration as the second element. The second element can be of any type, "+
                             "including `None` in case there is no particular information.")

```

tests/test_implicit_priors.py
```diff
@@ -38,6 +38,32 @@ def func(x, restoration_strength=0.1):
...
+        return x, None, False
+    restore_prior_3 = cuqi.implicitprior.RestorationPrior(func_3)
+    with pytest.raises(ValueError, match=r"Wrong return type .*"):
+        restore_prior_3.restore(np.ones(4), 0.1)
+    # Invalid return type 4: list with 2 elements
+    def func_4(x, restoration_strength=0.1):
+        return [x, None]
+    restore_prior_4 = cuqi.implicitprior.RestorationPrior(func_4)
+    with pytest.raises(ValueError, match=r"Wrong return type .*"):
+        restore_prior_4.restore(np.ones(4), 0.1)
```
amal-ghamdi, line 66: Nice tests! 

Just wanted to mention if you accept the updates to the error msg above, the `match` string needs to be updated.

amal-ghamdi (commented):
Thank you @chaozg for addressing this issue! I like the solution.

I only left few comments about some text (doc, error msg).

amal-ghamdi (approved):
LGTM!

PR #602, opened 2025-01-07
Use scipy.optimize.lsq_linear in RegularizedLinearRTO for nonnegativity

### LINKED ISSUE #596: Add option to use scipy.optimize.least_squares in RegularizedLinearRTO

chaozg:
fixed #596


cuqi/experimental/mcmc/_rto.py
```diff
@@ -215,9 +218,18 @@ def __init__(self, target=None, initial_point=None, maxit=100, inner_max_it=10,
         self.maxit = maxit
         self.inner_max_it = inner_max_it
         self.penalty_parameter = penalty_parameter
+        self.inner_solver = inner_solver
 
     def _initialize(self):
         super()._initialize()
+        if self.inner_solver == "LSQ":
+            self._inner_solver = self.inner_solver
```
jeverink, line 226: Please add the solver to the list of solvers in the docstring of the class.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -196,6 +196,9 @@ class RegularizedLinearRTO(LinearRTO):
     
     adaptive : bool
         If True, FISTA is used as inner solver, otherwise ISTA is used. *Optional*.
+    
+    inner_solver : string
+        If set to "LSQ", the inner solver is set to cuqi.solver.ScipyLinearLeastSquares. *Optional*.
```
jeverink, line 201: LSQ (least-squares) might not be a good name for this solver, as all solvers solve regularized or constrained linear least-squares problems. For consistency, it is probably best to refer to the name of the actual algorithm used by the scipy solver. According to their documention, it is a particular interior-point-like method called the Trust Region Reflective (trf) algorithm.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -215,9 +218,18 @@ def __init__(self, target=None, initial_point=None, maxit=100, inner_max_it=10,
...
         self.inner_max_it = inner_max_it
         self.penalty_parameter = penalty_parameter
+        self.inner_solver = inner_solver
 
     def _initialize(self):
         super()._initialize()
+        if self.inner_solver == "LSQ":
+            self._inner_solver = self.inner_solver
+            if self.target.prior._preset == "nonnegativity":
+                self._box_bounds = (np.ones(self.prior.dim)*0, np.ones(self.prior.dim)*np.inf)
```
jeverink, line 228: Should this be done here, or would it be better for the `_box_bounds` to be constructed in the RegularizedGaussian class and then retrieved here?

jeverink (changes requested):
Looks good, but my main issue is the use of name LSQ for the solver.

cuqi/experimental/mcmc/_rto.py
nabriis: What is the default choice? Unclear from this docstring. I suggest listing the solver options.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -259,7 +271,14 @@ def step(self):
...
         elif self._inner_solver == "ADMM":
             sim = ADMM(self.M, y, self.proximal,
-                        self.current_point, self.penalty_parameter, maxit = self.maxit, inner_max_it = self.inner_max_it, adaptive = self.adaptive)  
+                        self.current_point, self.penalty_parameter, maxit = self.maxit, inner_max_it = self.inner_max_it, adaptive = self.adaptive)
+        elif self._inner_solver == "LSQ":
+                A_op = sp.sparse.linalg.LinearOperator((sum([llh.dim for llh in self.likelihoods])+self.prior.dim, self.prior.dim),
+                                        matvec=lambda x: self.M(x, 1),
+                                        rmatvec=lambda x: self.M(x, 2)
+                                        )
+                print("Using ScipyLinearLeastSquares")
```
nabriis, line 280: What is this print doing here?

nabriis (changes requested):
Thanks @chaozg. I only had a few comments. Overall looks good!

cuqi/experimental/mcmc/_rto.py
```diff
@@ -259,7 +270,17 @@ def step(self):
...
         elif self._inner_solver == "ADMM":
             sim = ADMM(self.M, y, self.proximal,
-                        self.current_point, self.penalty_parameter, maxit = self.maxit, inner_max_it = self.inner_max_it, adaptive = self.adaptive)  
+                        self.current_point, self.penalty_parameter, maxit = self.maxit, inner_max_it = self.inner_max_it, adaptive = self.adaptive)
+        elif self._inner_solver == "ScipyLinearLSQ":
+                A_op = sp.sparse.linalg.LinearOperator((sum([llh.dim for llh in self.likelihoods])+self.target.prior.dim, self.target.prior.dim),
+                                        matvec=lambda x: self.M(x, 1),
+                                        rmatvec=lambda x: self.M(x, 2)
+                                        )
+                sim = ScipyLinearLSQ(A_op, y, self.target.prior._box_bounds,
```
amal-ghamdi, line 279: since we do not use self.proximal here, I suggest adding a mechanism to detect if there is unsupported proximal in the prior. For example, check `self.prior._preset` is either 'nonnegativity' or 'box' and raise an error otherwise

cuqi/experimental/mcmc/_rto.py
chaozg: Thanks @jeverink for your consideration here. Being specific about which algorithm is being used indeed has its advantages. Here we consider LSQ as a "solver" in the sense that it solves our constrained optimization problem. To address this issue, I just updated the docstring to make it clear that it's "Wrapper for Scipy's lsq_linear for the Trust Region Reflective algorithm".
> LSQ (least-squares) might not be a good name for this solver, as all solvers solve regularized or constrained linear least-squares problems. For consistency, it is probably best to refer to the name of the actual algorithm used by the scipy solver. According to their documention, it is a particular interior-point-like method called the Trust Region Reflective (trf) algorithm.

cuqi/experimental/mcmc/_rto.py
chaozg: Yes, suggestion accepted! I moved the defition of `_box_bounds` to `RegularizedGaussian `.

cuqi/experimental/mcmc/_rto.py
chaozg: It's part of my debug code, and I forgot to remove it : )  Removed now

tests/test_solver.py
```diff
@@ -54,28 +54,66 @@ def test_ScipyLBFGSB_with_gradient():
...
 
+def test_ScipyLinearLSQ_with_matrix():
+    rng = np.random.default_rng(seed = 1219)
+    m, n = 10, 5
+    A = rng.standard_normal((m, n))
+    b = rng.standard_normal(m)
+    res = lsq_linear(A, b, tol=1e-8)
+    ref_sol = res.x
+    sol, _ = ScipyLinearLSQ(A, b).solve()
+    assert np.allclose(sol, ref_sol, rtol=1e-3)
```
amal-ghamdi, line 87: just curious, in this test and the next, can we tighten the rtol to for example 1e-10, since it is the same underlying algorithm? (might need to fix the seed before each to use the same seed before each solve)

tests/test_solver.py
```diff
@@ -54,28 +54,66 @@ def test_ScipyLBFGSB_with_gradient():
...
+    # solve with ScipyLinearLSQ
+    lb = np.zeros(3)
+    ub = lb + np.inf
+    sol_lsq, _ = ScipyLinearLSQ(A, b, (lb,ub)).solve()
+    # solve with FISTA
+    rng = np.random.default_rng(seed = 1219)
+    x0 = rng.standard_normal(3)
+    sol_fista, _ = FISTA(A, b, lambda x, _: ProjectNonnegative(x), x0, stepsize=1e-7, maxit=100000, abstol=1e-16, adaptive=True).solve()
+
+    assert np.allclose(sol_lsq, sol_fista, rtol=1e-3)
```
amal-ghamdi, line 115: cool  😎

cuqi/experimental/mcmc/_rto.py
chaozg: Thanks @nabriis . Is the following better now?

`If set to "ScipyLinearLSQ", solver is set to cuqi.solver.ScipyLinearLSQ, otherwise FISTA/ISTA is used. Note "ScipyLinearLSQ" can only be used with `RegularizedGaussian` of `box` or `nonnegativity` constraint.`

> What is the default choice? Unclear from this docstring. I suggest listing the solver options.

amal-ghamdi:
Thank you @chaozg for this PR! It's great to add an efficient option for the solver!

I added a few comments, I noticed you made updates at the same time I was writing my review. So if there is anything from my comments that is no longer relevant, please just ignore it.

cuqi/experimental/mcmc/_rto.py
chaozg: Hi @amal-ghamdi, do you mean incompatible solver and constraint? I just added the following check in `_initialize()`, so `_solver` will only be set to `ScipyLinearLSQ ` if the constraint type is `box` or `nonnegativity `.
```python
if self.solver == "ScipyLinearLSQ":
    if (self.target.prior._preset == "nonnegativity" or self.target.prior._preset == "box"):
        self._solver = self.solver
    else:
        raise ValueError("ScipyLinearLSQ can only be used with RegularizedGaussian of box or nonnegativity constraint.")
```
Is this something close to what you suggested?

cuqi/solver/_solver.py
```diff
@@ -227,6 +227,43 @@ def solve(self):
...
+    def solve(self):
+        """Runs optimization algorithm and returns solution and optimization result.
+
+        Returns
+        ----------
+        solution : Tuple
+            Solution found (array_like) and optimization result (OptimizeResult).
+        """
+        res = opt.lsq_linear(self.A, self.b, bounds=self.bounds, **self.kwargs)
+        return res.x, res
```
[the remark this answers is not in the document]
chaozg: Done

cuqi/experimental/mcmc/_rto.py
chaozg: > Also it is not clear to me why to say inner solver. can we just call it solver

Yes, "solver" is better, particularly because, within linear LSQ, there is an internal solver to solve the unconstrained optimization .

cuqi/experimental/mcmc/_rto.py
```diff
@@ -204,21 +211,28 @@ class RegularizedLinearRTO(LinearRTO):
...
+        self.solver = solver
 
     def _initialize(self):
         super()._initialize()
-        if self._inner_solver == "FISTA":
+        if self.solver == "ScipyLinearLSQ":
+            if (self.target.prior._preset == "nonnegativity" or self.target.prior._preset == "box"):
+                self._solver = self.solver
+            else:
+                raise ValueError("ScipyLinearLSQ can only be used with RegularizedGaussian of box or nonnegativity constraint.")
```
jeverink, line 234:
```suggestion
                raise ValueError("ScipyLinearLSQ only supports RegularizedGaussian with box or nonnegativity constraint.")
```

jeverink (approved):
LGTM

cuqi/experimental/mcmc/_rto.py
```diff
@@ -188,14 +189,20 @@ class RegularizedLinearRTO(LinearRTO):
...
+    
+    inner_abstol : float
+        Tolerance parameter for ScipyLinearLSQ's inner solve of the unbounded least-squares problem. *Optional*.
     
     adaptive : bool
-        If True, FISTA is used as inner solver, otherwise ISTA is used. *Optional*.
+        If True, FISTA is used as solver, otherwise ISTA is used. *Optional*.
+    
+    solver : string
+        If set to "ScipyLinearLSQ", solver is set to cuqi.solver.ScipyLinearLSQ, otherwise FISTA/ISTA is used. Note "ScipyLinearLSQ" can only be used with `RegularizedGaussian` of `box` or `nonnegativity` constraint. *Optional*.
```
amal-ghamdi, line 205: Do we need to say ADMM here?
```suggestion
        If set to "ScipyLinearLSQ", solver is set to cuqi.solver.ScipyLinearLSQ, otherwise FISTA/ISTA or ADMM is used. Note "ScipyLinearLSQ" can only be used with `RegularizedGaussian` of `box` or `nonnegativity` constraint. *Optional*.
```

cuqi/experimental/mcmc/_rto.py
```diff
@@ -204,21 +211,28 @@ class RegularizedLinearRTO(LinearRTO):
...
 
     def _initialize(self):
         super()._initialize()
-        if self._inner_solver == "FISTA":
+        if self.solver == "ScipyLinearLSQ":
+            if (self.target.prior._preset == "nonnegativity" or self.target.prior._preset == "box"):
+                self._solver = self.solver
+            else:
+                raise ValueError("ScipyLinearLSQ only supports RegularizedGaussian with box or nonnegativity constraint.")
+        if self._solver == "FISTA":
```
amal-ghamdi, line 235: I think this is a nice error check. One suggestion is instead of using self.solver and self._solver is to make solver a property that does the check (and it also does the job of line 246),

something like (note this code is just for illustration and I did not test it :))
```python
@property
def solver(self):
    return self._solver
@solver.setter
def solver(self, value):
    if value == "ScipyLinearLSQ":
        if (self.target.prior._preset == "nonnegativity" or self.target.prior._preset == "box"):
            self._solver = value
        else:
            raise ValueError("ScipyLinearLSQ only supports RegularizedGaussian with box or nonnegativity constraint.") 
    else:
        self._solver = "FISTA" if callable(self.proximal) else "ADMM"
```

amal-ghamdi (approved):
Many thanks @chaozg  for the updates! I have only two comments about adding a solver property and small documentation update for your consideration. I approved.

cuqi/experimental/mcmc/_rto.py
amal-ghamdi: Yes, perfect. I just added a comment about whether this better fits in a property.

cuqi/experimental/mcmc/_rto.py
chaozg: Thanks @amal-ghamdi. Your suggested code is integrated with tiny changes : )

amal-ghamdi: looks good, thanks @chaozg !

nabriis (approved):

PR #610, opened 2025-01-17
Multiple Inputs: update the Model module

### LINKED ISSUE #605: Multiple Inputs: Update forward model to accept multiple inputs

This is the second task outlined in the [design document](https://github.com/CUQI-DTU/CUQIpy-admin/blob/multiple_input_design/design/fwd_multiple_input.ipynb) for enabling multiple inputs. This task is to update the `model` module classes to enable multiple inputs for the forward map. It also address enabling gradient with respect to multiple inputs.

- [x] Classes of the `model` module are updated to support multiple inputs
- [x] (except PDE models to be differed to future issue) Classes of the `model` module are updated  to support gradient computation with respect to multiple inputs
- [x] (deferred to future issue due to the size of this PR) Raise error when model with multiple input is used within distributions (e.g. in the likelihood) since this is to be supported in upcoming tasks #649
- [x] (if applicable) The design file linked above is updated, this sub-feature use and functionality (it was updated).

- [x] Unit tests added/updated (and passing!) about 400 test cases added

- [x] (canceled, deferred for later tasks in the multiple input feature) HowTo/tutorial added

**additional notes**
Notes about the user interface and enabled features: please consult  [design document](https://github.com/CUQI-DTU/CUQIpy-admin/blob/multiple_input_design/design/fwd_multiple_input.ipynb) for the most recent added features
```python
## Forward map
# consider the forward model F we defined above
F(a_val, b_val, c_val) # should work (if passed as args, input should be in the same order 
                       # as what the domain geometry specifies, i.e. a, b, c)
                       # when possible, error is raised if the order is not correct
                       # this type of error can be detected, for example, when the 
                       # inputs a, b, and c have different parameter dimensions
F(a=a_val, b=b_val, c=c_val) # should work
F(b=b_val, c=c_val, a=a_val) # should work
# Future: support partial evaluation
F(a=a_val) # should work in the future, but not part of this task and returns
           # another forward model that is partially evaluated
#---------------------------------------------------
## Gradient
F.gradient(direction, a_val, b_val, c_val) # should work
```
~~# when possible also combination like below should work~~
~~F.gradient(direction, a=a_val) # should work~~
~~F.gradient(direction, a=a_val, c=c_val) # should work~~

**Notes about reviewing**
- I suggest looking at the [design document](https://github.com/CUQI-DTU/CUQIpy-admin/blob/multiple_input_design/design/fwd_multiple_input.ipynb) first, which I updated. The relevant part, is from the beginning up to the heading "The rest of the document is not yet updated". The content below this heading is not relevant to this review.
- The comparison file between Model class before and after the update helps identifying the methods that were added, [model_update.md](https://github.com/CUQI-DTU/CUQIpy-admin/blob/multiple_input_design/design/model_update.md).
- It might be better to review the tests before the model file because they showcase how to use the feature and what scenarios generate errors
- For the Model class, it is better to review the entire class (also referring to the diff file). Reviewing the diff file only can be confusing due to the so many changes
- For the remaining classes, LinearModel, PDEModel, etc., it makes sense to focus on the diff file only
- There are some remaining issues I will list them here: #648 #649

amal-ghamdi:
closes #605 

Notes about reviewing: see issue description


nabriis (commented):
**From live review of code**
1) Printing looks a bit strange for Model. Maybe use specific case for ProductGeometry?
2) Can you check that if we pass in CUQIArray but say its funval/parameters what happens? Works as intended?
3) Related to stacking. Do we have good docstring to explain this option? Should we ask for user to provide e.g. "is_stacked".
4) What about Samples in the Dict? Why is the stacked samples not support?
5) Check that given RVs x, y, z, that F(x, y, z) works.
6) Look into gradient as None. Do we need  it now? Seems to change behavior. Maybe we return error?
7) Gradient as stacked. Look into "is_stacked" parameter.
8) Look into stacked output.
9) Look into if Model should be two classes or one?

amal-ghamdi:
@nabriis , if you got a chance, it will be great if you can provide your review on the PR. I am also happy to go over parts that are not clear in a meeting :). Many thanks!

cuqi/model/_model.py
```diff
@@ -87,64 +92,140 @@ def jacobian(x): # Can use "x" or "wrt" as the input argument name
...
+    .. code-block:: python
+
+        import numpy as np
+        from cuqi.model import Model
+        from cuqi.geometry import Discrete
+    
+        def forward(x, y):
+            return 10 * y - 10 * x**3 + 5 * x**2 + 6 * x
+    
+        def jacobian(x, y):
```
chaozg, line 117: What is the signature of the gradient function like?

cuqi/model/_model.py
```diff
@@ -160,334 +241,808 @@ def range_dim(self):
...
+            not isinstance(
+                self.domain_geometry, cuqi.experimental.geometry._ProductGeometry
+            )
+            and len(non_default_args) > 1
+        ):
+            raise ValueError(
+                "The forward operator input is specified by more than one argument. This is only supported for domain geometry of type tuple of cuqi.geometry.Geometry objects."
+            )
+
+    def _check_correct_gradient_jacobian_form(self, func, func_type):
```
chaozg, line 259: Beside signature, I wonder if it makes sense to also check the return type and possibly size.

cuqi/model/_model.py
```diff
@@ -160,334 +241,808 @@ def range_dim(self):
...
+            return all(
+                d.dim == self.domain_geometry.par_dim_list[i]
+                for i, d in enumerate(distributions)
+            )
+        else:
+            return False
 
-    def __call__(self, *args, **kwargs):
-        return self.forward(*args, **kwargs)
+    def _handle_distributions(self, kwargs):
```
chaozg, line 769: Could we have a more expressive name? I mean it's not clear what this function is doing from its name (even with together with the docstring)

chaozg (commented):
Tiny comments on `init` and `forward`

cuqi/model/_model.py
```diff
@@ -87,64 +92,140 @@ def jacobian(x): # Can use "x" or "wrt" as the input argument name
...
+    def domain_geometry(self, value):
+        """ Update the domain geometry of the model. """
+        # Forward model with multiple input case
+        if isinstance(value, tuple) and\
+            all([isinstance(v, Geometry) for v in value]):
+            self._domain_geometry = cuqi.experimental.geometry._ProductGeometry(
+                *value)
+        # Remaining cases, forward model with single input
+        elif isinstance(value, tuple) and len(value) == 2:
+            self._domain_geometry = _DefaultGeometry2D(value)
```
nabriis, line 216: This may need an overhaul depending if single or multiple inputs

cuqi/model/_model.py
```diff
@@ -160,334 +241,808 @@ def range_dim(self):
...
-        """ Converts `x` to function values (if needed) using the appropriate 
-        geometry. For example, `x` can be the model input which need to be
-        converted to function value before being passed to 
-        :class:`~cuqi.model.Model` operators (e.g. _forward_func, _adjoint_func,
-        _gradient_func).
+    def _check_domain_geometry_consistent_with_forward(self):
+        """Private function that checks if the domain geometry of the model is
+        consistent with the forward operator."""
+
+        non_default_args = cuqi.utilities.get_non_default_args(self._forward_func)
```
nabriis, line 248: We have a property for this. Why repeat it here?

cuqi/model/_model.py
```diff
@@ -160,334 +241,808 @@ def range_dim(self):
...
+        if (
+            not isinstance(
+                self.domain_geometry, cuqi.experimental.geometry._ProductGeometry
+            )
+            and len(non_default_args) > 1
+        ):
+            raise ValueError(
+                "The forward operator input is specified by more than one argument. This is only supported for domain geometry of type tuple of cuqi.geometry.Geometry objects."
+            )
```
nabriis, line 257: This is rather besides the point of this PR, but the formatting here using (i think) black makes the review somewhat harder in my opinion.

cuqi/model/_model.py
```diff
@@ -160,334 +241,808 @@ def range_dim(self):
...
+        args = list(kwargs.values())
+        if len(args)== 1 and isinstance(args[0], cuqi.experimental.algebra.RandomVariable):
+            return self._handle_random_variable(args[0])
+
         # If input is a Node from internal abstract syntax tree, we let the Node handle the operation
         # We use NotImplemented to indicate that the operation is not supported from the Model class
         # in case of operations such as "@" that can be interpreted as both __matmul__ and __rmatmul__
         # the operation may be delegated to the Node class.
-        if isinstance(x, cuqi.experimental.algebra.Node):
+        if len(args)== 1 and isinstance(args[0], cuqi.experimental.algebra.Node):
```
nabriis, line 741:
```suggestion
        if len(args) == 1 and isinstance(args[0], cuqi.experimental.algebra.Node):
```

cuqi/model/_model.py
```diff
@@ -160,334 +241,808 @@ def range_dim(self):
...
+               for x in kwargs.values()):
+            return self._handle_distributions(kwargs)
 
         # If input is a random variable, we handle it separately
-        if isinstance(x, cuqi.experimental.algebra.RandomVariable):
-            return self._handle_random_variable(x)
-        
+        # extract args from kwargs
+        args = list(kwargs.values())
+        if len(args)== 1 and isinstance(args[0], cuqi.experimental.algebra.RandomVariable):
```
nabriis, line 734:
```suggestion
        if len(args) == 1 and isinstance(args[0], cuqi.experimental.algebra.RandomVariable):
```

nabriis (commented):
Thanks @amal-ghamdi. I found some time to do a review over the holiday. I did not dive into the stacked options though.

Overall the feature seems very close to complete. The comprehensive testing framework is really nice! I had some small observations + the following comments:


1) I suggest updating the repr of model now that we can have geometries that can span multiple lines. To avoid this:
```bash
CUQI Model: _ProductGeometry(
	Discrete(1,)
	Discrete(1,)
) -> _DefaultGeometry1D(1,).
    Forward parameters: ['x', 'y'].
```
2) There seems to be an input discrepancy we need to look into. For models with multiple inputs one might expect to be able to do this:
```python
def forward(x, y):
    return 10 * y - 10 * x**3 + 5 * x**2 + 6 * x

def jacobian(x, y):
    return (np.array([[-30 * x**2 + 10 * x + 6]]), np.array([[10]]))

model = Model(
    forward,
    range_geometry=1,
    domain_geometry=(1, 1),
    jacobian=jacobian,
)
```
however this is not possible.
3) For this case, I notice the behavior is different than for 1 variable. Namely, we "lost" the model information because the parameter names should only have been changed.
```python
import numpy as np
import cuqi
from cuqi.model import Model
from cuqi.geometry import Discrete

def forward(x, y):
    return 10 * y - 10 * x**3 + 5 * x**2 + 6 * x

def jacobian(x, y):
    return (np.array([[-30 * x**2 + 10 * x + 6]]), np.array([[10]]))

model = Model(
    forward,
    range_geometry=1,
    domain_geometry=(Discrete(1), Discrete(1)),
    jacobian=jacobian,
)

z = cuqi.experimental.algebra.RandomVariable(cuqi.distribution.Normal(0,1))
v = cuqi.experimental.algebra.RandomVariable(cuqi.distribution.Normal(0,1))

model(z,v)
```

cuqi/model/_model.py
```diff
@@ -160,334 +241,808 @@ def range_dim(self):
...
+        geometries, is_par = self._helper_pre_conversion_checks_and_processing(geometry, is_par, **kwargs)
+
+        # Convert to function values
+        for i, (k, v) in enumerate(kwargs.items()):
+            # Use CUQIarray funvals if geometry is consistent
+            if isinstance(v, CUQIarray) and v.geometry == geometries[i]:
+                kwargs[k] = v.funvals
+            # Otherwise we use the geometry par2fun method
+            elif is_par[i] and v is not None:
+                kwargs[k] = geometries[i].par2fun(v)
```
chaozg, line 406: The comments seem not sync with the code. I mean 'Otherwise' with `elif`.

cuqi/model/_model.py
amal-ghamdi: I updated the name, let me know if you have other suggestions.

cuqi/model/_model.py
```diff
@@ -160,334 +241,808 @@ def range_dim(self):
...
+        out = func(**kwargs)
 
-        if len(args) > 0:
+        # Return output as parameters
+        # (wrapped in CUQIarray if any input was CUQIarray)
+        return self._2par(
+            geometry=func_range_geometry, to_CUQIarray=is_CUQIarray, **{"out": out}
+        )["out"]
+
+    def _handle_samples(self, func=None, fwd=True, **kwargs):
```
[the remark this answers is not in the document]
amal-ghamdi: I also updated the name here

cuqi/model/_model.py
amal-ghamdi: I made few updates to clarify this

amal-ghamdi:
> Thanks @amal-ghamdi. I found some time to do a review over the holiday. I did not dive into the stacked options though.
> 
> Overall the feature seems very close to complete. The comprehensive testing framework is really nice! I had some small observations + the following comments:
> 
> 1. I suggest updating the repr of model now that we can have geometries that can span multiple lines. To avoid this:
> 
> ```shell
> CUQI Model: _ProductGeometry(
> 	Discrete(1,)
> 	Discrete(1,)
> ) -> _DefaultGeometry1D(1,).
>     Forward parameters: ['x', 'y'].
> ```
> 
> 2. There seems to be an input discrepancy we need to look into. For models with multiple inputs one might expect to be able to do this:
> 
> ```python
> def forward(x, y):
>     return 10 * y - 10 * x**3 + 5 * x**2 + 6 * x
> 
> def jacobian(x, y):
>     return (np.array([[-30 * x**2 + 10 * x + 6]]), np.array([[10]]))
> 
> model = Model(
>     forward,
>     range_geometry=1,
>     domain_geometry=(1, 1),
>     jacobian=jacobian,
> )
> ```
> 
> however this is not possible. 3) For this case, I notice the behavior is different than for 1 variable. Namely, we "lost" the model information because the parameter names should only have been changed.
> 
> ```python
> import numpy as np
> import cuqi
> from cuqi.model import Model
> from cuqi.geometry import Discrete
> 
> def forward(x, y):
>     return 10 * y - 10 * x**3 + 5 * x**2 + 6 * x
> 
> def jacobian(x, y):
>     return (np.array([[-30 * x**2 + 10 * x + 6]]), np.array([[10]]))
> 
> model = Model(
>     forward,
>     range_geometry=1,
>     domain_geometry=(Discrete(1), Discrete(1)),
>     jacobian=jacobian,
> )
> 
> z = cuqi.experimental.algebra.RandomVariable(cuqi.distribution.Normal(0,1))
> v = cuqi.experimental.algebra.RandomVariable(cuqi.distribution.Normal(0,1))
> 
> model(z,v)
> ```


Really good points, 

**regarding the random variables**, when I finished this PR, I meant to add this issue as a separate issue but did not document that in here (only documented in personal notes). So now I just fixed it here with this PR instead, which makes sense. 

**I also updated the accepted domain geometry type** to support the example you provided along with updating the documentation of domain and range geometry. You can see the documentation and the added test `test_setting_domain_and_range_geometry` in `test_model.py` for the accepted format

**I updated the formatting**, single input:
```
CUQI Model: _DefaultGeometry1D(1,) -> _DefaultGeometry1D(1,).
    Forward parameters: ['x'].
```

Multiple inputs:
```
CUQI Model: _ProductGeometry(
                _DefaultGeometry1D(1,)
                _DefaultGeometry1D(1,)
            ) -> _DefaultGeometry1D(1,).
    Forward parameters: ['x', 'y'].
```

cuqi/model/_model.py
amal-ghamdi: updated as suggested through a new property `number_of_inputs`

cuqi/model/_model.py
amal-ghamdi: I know 🙂, I left it as is for now which is auto-formatted.

amal-ghamdi:
Many thanks @chaozg and @nabriis for your review! I know this is unusually big PR, so many thanks for taking the time to go through it. I addressed your points and I am requesting your review again.

@nabriis 
I added a generalization for the case of random variable internal nodes (from one input to multiple input). Let me know if this makes sence (line 771 in model):
```
        # the operation may be delegated to the Node class.
        elif any(isinstance(args_i, cuqi.experimental.algebra.Node) for args_i in args):
            return NotImplemented
```

nabriis:
> Many thanks @chaozg and @nabriis for your review! I know this is unusually big PR, so many thanks for taking the time to go through it. I addressed your points and I am requesting your review again.
> 
> @nabriis I added a generalization for the case of random variable internal nodes (from one input to multiple input). Let me know if this makes sence (line 771 in model):
> 
> ```
>         # the operation may be delegated to the Node class.
>         elif any(isinstance(args_i, cuqi.experimental.algebra.Node) for args_i in args):
>             return NotImplemented
> ```

Thanks @amal-ghamdi. Yes I think this is the right update.

nabriis (approved):
Thanks @amal-ghamdi  for reacting to my changes. The updates LGTM! Let us get this one merged ASAP!

cuqi/model/_model.py
```diff
@@ -160,341 +264,841 @@ def range_dim(self):
...
+                        [i_g.__name__ for i_g in _get_identity_geometries()],
+                    )
+                )
+
+    def _apply_chain_rule_to_account_for_domain_geometry_gradient(self,
+                                                                  kwargs_par,
+                                                                  grad,
+                                                                  grad_is_par,
+                                                                  to_CUQIarray):
+        """ Private function that applies the chain rule to account for the
```
chaozg, line 1043: It will be good to state/hint in the docstring or somewhere else, like next to corresponding lines of codes, that in the end we are computing the gradient with respect to the "parameter" values, not the "function" values.

cuqi/model/_model.py
```diff
@@ -599,19 +1249,35 @@ def shift(self, value):
...
+    def _forward_func_no_shift(self, *args, is_par=True, **kwargs):
+        """Helper function for computing the forward operator without the shift."""
+        # convert args to kwargs
+        kwargs = self._parse_args_add_to_kwargs(
+            *args, **kwargs, map_name="model", is_par=is_par
+        )
+        args = list(kwargs.values())
+        # if model has _original_non_default_args, we use it to replace the
+        # kwargs keys so that it matches self._linear_operator signature
+        if hasattr(self, '_original_non_default_args'):
```
chaozg, line 1261: Just a reminder from our discussion that in case this check is also needed in other functions

chaozg (commented):
@amal-ghamdi, I just left two comments from our review discussion

tests/test_model.py
```diff
@@ -1,11 +1,1323 @@
...
+        self.domain_geometry = None
+        self.range_geometry = None
+        self.test_data = []  # list of TestCase objects holding the test data
+        self.model_variations = (
+            []
+        )  # list of model variations which all share the same forward map
+        self.input_bounds = None  # bounds for the test input values
+
+    @property
+    def has_gradient(self):
```
chaozg, line 380: Could you add a bit comment on what this function is actually checking?

chaozg (approved):
LGTM. Thanks @amal-ghamdi .

cuqi/model/_model.py
amal-ghamdi: Thank you @chaozg, I doubled checked this case and realize we do not need to do something about it. Because changing the default argument names should only affect the parameter names for forward and gradient but not adjoint.

cuqi/model/_model.py
amal-ghamdi: done

tests/test_model.py
amal-ghamdi: done

PR #623, opened 2025-02-11
Enable callback in Gibbs and update callback interface to pass more information to the callback method

### LINKED ISSUE #622: Enable sampler callback in Gibbs and change interface of callback to pass more information

Enable sampler callback in Gibbs and enable callback to access more information, e.g. Ns, all samples, etc. This update is only in the experimental module and not breaking backward compatibility of the sampler module.

- [x] Change interface of callback to pass the sampler object itself and the sample_index and total number of samples instead of the current sample and the sample_index (note: we can also pass both sample object and current_sample)
- [x] Callback enabled in HybridGibbs
- [x] Update any documentation or demos that use callback (none use the experimental module callback)

- [x] Unit tests added/updated (and passing!)

- [x] (I suggest no need for now) HowTo/tutorial added

**Example**
```python
def callback(sampler, sample_index, num_of_samples)
    if sample_index %  num_of_sample == 100:
        # do something. e.g. save samples, plot samples, look at some diagnostics
        # something like this
        print(samples.compute_ESS()) 

```

amal-ghamdi:
closes #622

### DIFF
--- a/cuqi/experimental/mcmc/_gibbs.py
+++ b/cuqi/experimental/mcmc/_gibbs.py
@@ -59,7 +59,7 @@ class HybridGibbs:
         Default is 1 for all variables.
 
     callback : callable, optional
-        A function that will be called after each sampling step. The function should take two arguments: the sampler object and the index (integer) of the current sampling step. The callback function is useful for monitoring the sampler during sampling.
+        A function that will be called after each sampling step. The function should take three arguments: the sampler object, the index of the current sampling step, the total number of requested samples. The last two arguments are integers. The callback function is useful for monitoring the sampler during sampling.
 
     Example
     -------
@@ -164,13 +164,14 @@ class HybridGibbs:
             The number of samples to draw.
 
         """
-        self._Ns = Ns
         for idx in tqdm(range(Ns), "Sample: "):
 
             self.step()
-            self.callback(self, idx)
 
             self._store_samples()
+
+            # Call callback function if specified
+            self._call_callback(len(self._samples)-1, Ns)
 
         return self
 
@@ -198,6 +199,9 @@ class HybridGibbs:
                 self.tune(tune_interval, idx // tune_interval) 
                 
             self._store_samples()
+
+            # Call callback function if specified
+            self._call_callback(len(self._samples)-1, Nb)
 
         return self
 
@@ -270,6 +274,11 @@ class HybridGibbs:
             self.samplers[par_name].tune(skip_len=skip_len, update_count=update_count)
 
     # ------------ Private methods ------------
+    def _call_callback(self, sample_index, total_num_of_samples):
+        """ Calls the callback function. Assumes input is sampler, sample index, and total number of samples """
+        if self.callback is not None:
+            self.callback(self, sample_index, total_num_of_samples)
+
     def _initialize_samplers(self):
         """ Initialize samplers """
         for sampler in self.samplers.values():
--- a/cuqi/experimental/mcmc/_gibbs.py
+++ b/cuqi/experimental/mcmc/_gibbs.py
@@ -103,7 +103,7 @@ class HybridGibbs:
             
     """
 
-    def __init__(self, target: JointDistribution, sampling_strategy: Dict[str, Sampler], num_sampling_steps: Dict[str, int] = None):
+    def __init__(self, target: JointDistribution, sampling_strategy: Dict[str, Sampler], num_sampling_steps: Dict[str, int] = None, callback=None):
 
         # Store target and allow conditioning to reduce to a single density
         self.target = target() # Create a copy of target distribution (to avoid modifying the original)
@@ -119,6 +119,8 @@ class HybridGibbs:
 
         # Initialize sampler (after target is set)
         self._initialize()
+
+        self.callback = callback
 
     def _initialize(self):
         """ Initialize sampler """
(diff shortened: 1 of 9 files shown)


cuqi/experimental/mcmc/_gibbs.py
```diff
@@ -263,6 +275,11 @@ def tune(self, skip_len, update_count):
             self.samplers[par_name].tune(skip_len=skip_len, update_count=update_count)
 
     # ------------ Private methods ------------
+    def _call_callback(self, sample_index, num_of_samples):
```
chaozg, line 278: I have a question about callback's signature: I understand that we have `sample_index` here as `samples[sample_index]` might be the most used. But why do we need `num_of_samples` here? Could you remind me of the use cases?

tests/zexperimental/test_mcmc.py
```diff
@@ -1510,3 +1510,101 @@ def test_RegularizedLinearRTO_ScipyLinearLSQ_option_invalid():
...
+    the sampler passes the correct arguments to the callback function. """
+
+    Ns = 10
+    ideces = [] # list to store the indices of the samples
+
+    # Define the callback function
+    def callback_function(callback_sampler, sample_index, num_samples):
+        assert isinstance(callback_sampler, sampler.__class__)
+        assert num_samples == Ns
+        ideces.append(sample_index)
```
chaozg, line 1577: indices?
```suggestion
        indices.append(sample_index)
```

tests/zexperimental/test_mcmc.py
```diff
@@ -1510,3 +1510,101 @@ def test_RegularizedLinearRTO_ScipyLinearLSQ_option_invalid():
...
+    ),
+]
+
+@pytest.mark.parametrize("sampler", callback_testing_sampler_instances)
+def test_passing_callback_to_sampler(sampler):
+    """ Test that the callback function is called by the sampler and
+    the sampler passes the correct arguments to the callback function. """
+
+    Ns = 10
+    ideces = [] # list to store the indices of the samples
```
chaozg, line 1571:
```suggestion
    indices = [] # list to store the indices of the samples
```

tests/zexperimental/test_mcmc.py
```diff
@@ -1510,3 +1510,101 @@ def test_RegularizedLinearRTO_ScipyLinearLSQ_option_invalid():
...
+            sampling_strategy={
+                "x": cuqi.experimental.mcmc.NUTS(max_depth=7),
+                "s": cuqi.experimental.mcmc.Conjugate(),
+            },
+            callback=callback_function,
+        )
+
+    # Sample Ns warmup samples
+    my_sampler.warmup(Ns)
+    assert np.allclose(ideces, np.arange(Ns))
```
chaozg, line 1594:
```suggestion
    assert np.allclose(indices, np.arange(Ns))
```

tests/zexperimental/test_mcmc.py
```diff
@@ -1510,3 +1510,101 @@ def test_RegularizedLinearRTO_ScipyLinearLSQ_option_invalid():
...
+
+    # reset the indices
+    ideces = []
```
chaozg, line 1597:
```suggestion
    indices = []
```

tests/zexperimental/test_mcmc.py
```diff
@@ -1510,3 +1510,101 @@ def test_RegularizedLinearRTO_ScipyLinearLSQ_option_invalid():
...
+
+    # Sample Ns samples
+    my_sampler.sample(Ns)
+    assert np.allclose(ideces, np.arange(Ns))
```
chaozg, line 1601:
```suggestion
    assert np.allclose(indices, np.arange(Ns))
```

chaozg (commented):
Thank you @amal-ghamdi for this PR. I'm just left a question on the signature of the callback function.

jakobsj (changes requested):
Looks very good to me, only spotted a few typos and left a question.

cuqi/experimental/mcmc/_gibbs.py
amal-ghamdi: Thank you @chaozg for your review! In my case I found Ns very useful to determine the ratio of samples. For example, if I want to generate a plot every 10% of the samples or so, I need to know the Ns. And Ns it is not stored in the sampler object but passed to sample and warmup methods. So these two methods have access to it and can pass it to the callback.

cuqi/experimental/mcmc/_gibbs.py
```diff
@@ -192,6 +201,9 @@ def warmup(self, Nb, tune_freq=0.1) -> 'HybridGibbs':
                 
             self._store_samples()
 
+            # Call callback function if specified
+            self._call_callback(idx, Nb)
```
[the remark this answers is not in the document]
amal-ghamdi: Thank you Jakob for your review! I agree that one might want different behaviour for warm up and sample. This actually also crossed my mind working in this PR. I thought ideally, the user should be able to pass 2 callbacks, e.g. `NUTS(target, sample_callback=fun1, warmup_callback=fun2)` where either of them could be None. The user should be able to, also, pass the same function for both as well `NUTS(target, sample_callback=fun1, warmup_callback=fun1)`.

I suggest, because this is a separate and new feature beyond what we normally support and because it needs design decision considerations, that we create a different issue ticket for it to be addressed in the future. It will also make the scope of this PR smaller and hopefully get merged faster :).

amal-ghamdi:
Thank you @chaozg and @jakobsj for your timely review! 😃. I addressed your comments and asked for your review again.

cuqi/experimental/mcmc/_gibbs.py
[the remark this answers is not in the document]
jakobsj: Agreed, please add this as a possible future issue.

jakobsj (approved):
Thanks for updating, happy to approve!

chaozg (approved):
LGTM!

PR #625, opened 2025-02-11
Enable general observation function for PDE, and accessing original solution (or other observable quantities)

### LINKED ISSUE #620: Enable general observation function in PDE and accessing original solution

In some cases, we need to observe derived quantities from the PDE solution, which is a function of the solution and the parameter. This is supported in CUQIpy-FEniCS plug in but not in core CUQIpy

- [x] (DoD item updated) Enable specifying observation operator for core PDE classes (this is already enabled)
- [x] (DoD item updated) Enable accessing the original solution if needed (this is already enabled)
- [x] (DoD item updated) Add helper function for spatial gradient
- [x] extract time space observation which is implemented now in the time dependent PDE in some utility function so that it can be used by the user

- [x] Unit tests added/updated (and passing!)

- [x] (Updated existing tutorial) HowTo/tutorial added (@jakobsj this item is updated)

amal-ghamdi:
closes #620

### DIFF
--- a/cuqi/pde/_pde.py
+++ b/cuqi/pde/_pde.py
@@ -387,7 +387,7 @@ class TimeDependentLinearPDE(LinearPDE):
         return solution_obs
 
     def observe(self, solution):
-        """Apply observation operator to the solution. This include
+        """Apply observation operator to the solution. This includes
         interpolation to observation points (if different from the
         solution grid) then applying the observation map (if provided)."""
 


cuqi/pde/_observation_map.py
```diff
@@ -0,0 +1,20 @@
...
+        solution_obs[i, :] = ((sol[i, :] - sol[i+1, :])/
+                                       (grid[i] - grid[i+1]))
+    return solution_obs
+
+def _extract_spatial_temporal_obs(sol, grid, times, grid_obs, time_obs):
+    """Private function to extract solution at observation points in space and
+    time."""
+    # Interpolate solution in space and time to the observation
+    # time and space
+    solution_obs = scipy.interpolate.RectBivariateSpline(
```
chaozg: I'm curious if `scipy.interpolate.RectBivariateSpline` works on irregular geometries besides rectangles and unstructured grids/meshes

amal-ghamdi:
> Hi @amal-ghamdi , thanks for implementing this feature.
> 
> Do you plan to have more demos on clarifying how this observation function could be used? For example, I wonder with an `observation_map` like `lambda u, grid_obs: u**2`, how do we compute the jacobian? I mean could the gradient be computed automatically? Is the chain product broken or 'preserved'? I guess a more common observation would be `lambda u, grid:u[some indices]`. Can we also compute the gradient in this case and use ULA/NUTS?

Many thanks @chaozg for your review. I updated a howto (demos/howtos/TimeDependentLinearPDE.py) we already have to show some usage of the `observation_map`.  The gradient is not supported automatically, the gradient of the entire forward map can be provided by the user when creating the PDEModel object. More elegant or efficient features can be done (like providing gradient for observation operator), but, I believe, it is not within the scope of this PR.

On the posterior level, one can enable FD gradient, regardless of the underlying model.

Issue #505, opened 2024-09-04
Expand BayesianProblem sampling Gibbs strategies when the experimental is set to True, to use for example NUTS, etc


amal-ghamdi:
We can create an issue to expand sampling strategies when the experimental is set to True, to use for example NUTS, etc

(likely after #503 is done and once we are sure NUTS/MH works well within Gibbs)

_Originally posted by @amal-ghamdi in https://github.com/CUQI-DTU/CUQIpy/pull/500#discussion_r1742271981_

Issue #195, opened 2023-02-22
Consolidate BayesianProblem, Posterior, JointDistribution etc.


nabriis:

amal-ghamdi:
should user only use BP and extract JointDistribution from the BP

Issue #95, opened 2022-09-16
Update BayesianProblem with new sampler choices


nabriis:
After #34 is implemented, it should be fairly easy to add more automatic sampler choices to BayesianProblem.

Some common choices would be

- [ ] NUTS (when gradient exists for that parameter)
- [ ] Conjugate (for InverseGamma + Gaussian)
- [ ] Direct (when the distribution can be directly sampled)
- [ ] CWMH (for small dimensional < 10) posteriors
- [ ] pCN (Gaussian prior, but non-linear model)
- [ ] Metropolis (Fallback, only for very small dim problems

PR #630, opened 2025-02-21
Robust automatic sampler selection/suggestion

jeverink:
This PR implements a robust method for automatically selecting and instantiating a sampler given a distribution, when possible.
It makes a suggestion based a predefined ordered list of sampler classes and finds the first valid ones. (Also works with HybridGibbs.)

This could be used to rework `sample_posterior` from `BayesianProblem` (only for the experimental samplers) to make it more maintainable by removing long if/else-chains. Could also help with Issues #505 #195 #95 

Called it `suggest_sampler` to insinuate that it is only a suggestion and might not be the best sampler for the given distribution.

Example:
image

The created sampler is a HybridGibbs sampler with Conjugacy and UGLA.

To-do's:
- [ ] Write an issue with DoD
- [ ] Discuss ordering of samplers
- [ ] Add conditions such that some samplers are perhaps only suggested on sufficiently small problems.
- [ ] Tests

### DIFF
--- a/cuqi/experimental/mcmc/_utilities.py
+++ b/cuqi/experimental/mcmc/_utilities.py
@@ -88,3 +88,28 @@ def suggest_sampler(target, as_string = True):
                 return suggestion
             
     return None
+
+def suggest_sampling_strategy(target):
+    """
+        Suggests a possible sampling strategy to be used with the HybridGibbs sampler.
+
+        For reasoning behind the suggestion, see 'suggest_sampler'.
+    """
+
+    if not isinstance(target, cuqi.distribution.JointDistribution):
+        return None
+
+    par_names = target.get_parameter_names()
+
+    suggested_samplers = dict()
+    for par_name in par_names:
+        conditional_params = {par_name_: np.ones(target.dim[i]) for i, par_name_ in enumerate(par_names) if par_name_ != par_name}
+        conditional = target(**conditional_params)
+
+        sampler = suggest_sampler(conditional, as_string = True)
+        if sampler is None:
+            return None
+        
+        suggested_samplers[par_name] = sampler
+
+    return suggested_samplers
--- a/cuqi/experimental/mcmc/_utilities.py
+++ b/cuqi/experimental/mcmc/_utilities.py
@@ -47,49 +47,57 @@ def find_valid_sampling_strategy(target, as_string = True):
 
     return valid_samplers
 
-def suggest_sampler(target, as_string = True):
+def suggest_sampler(target, as_string = False, exceptions = []):
     """
         Suggests a possible sampler that can be used for sampling from the target distribution.
 
         DESCRIBE ARGUMENTS BEHIND SUGGESTION
     """
 
+    n = target.dim
+
+    # Samplers with suggested default values (when no defaults are defined)
     ordering = [
         # Direct and Conjugate samplers
-        samplers.Direct,
-        samplers.Conjugate,
-        samplers.ConjugateApprox,
+        (samplers.Direct, {}),
+        (samplers.Conjugate, {}),
+        (samplers.ConjugateApprox, {}),
         # Specialized samplers
-        samplers.LinearRTO,
-        samplers.RegularizedLinearRTO,
-        samplers.UGLA,
+        (samplers.LinearRTO, {}),
+        (samplers.RegularizedLinearRTO, {}),
+        (samplers.UGLA, {}),
         # Hamiltonian samplers
-        samplers.NUTS,
+        (samplers.NUTS, {}),
         # Langevin based samplers
-        samplers.MALA,
-        samplers.ULA,
+        (samplers.MALA, {}),
+        (samplers.ULA, {}),
         # Gibbs and Componentwise samplers
-        samplers.HybridGibbs,
-        samplers.CWMH,
+        (samplers.HybridGibbs, {}),
+        (samplers.CWMH, {"scale" : 0.05*np.ones(n),
+                         "x0" : 0.5*np.ones(n)}),
         # Proposal based samplers
-        samplers.PCN,
-        samplers.MH,
+        (samplers.PCN, {"scale" : 0.02}),
+        (samplers.MH, {}),
     ]
 
     valid_samplers = find_valid_samplers(target, as_string = False)
     
-    for suggestion in ordering:
-        if suggestion in valid_samplers:
+    for suggestion, values in ordering:
+        if suggestion in valid_samplers and suggestion not in exceptions:
             # Sampler found
             if as_string:
                 return suggestion.__name__
-            else: 
-                # TODO: Instantiate if possible
-                return suggestion
+            else:
+                # Cases for samplers that might need default values
+                if suggestion is not samplers.HybridGibbs:
+                    return suggestion(target, **values)
+                else:
+                    return suggestion(target, sampling_strategy = suggest_sampling_strategy(target, as_string = False))
+
             
     return None
 
-def suggest_sampling_strategy(target):
+def suggest_sampling_strategy(target, as_string = False):
     """
         Suggests a possible sampling strategy to be used with the HybridGibbs sampler.
 
@@ -106,7 +114,7 @@ def suggest_sampling_strategy(target):
         conditional_params = {par_name_: np.ones(target.dim[i]) for i, par_name_ in enumerate(par_names) if par_name_ != par_name}
         conditional = target(**conditional_params)
 
-        sampler = suggest_sampler(conditional, as_string = True)
+        sampler = suggest_sampler(conditional, as_string = as_string)
         if sampler is None:
             return None
         
--- a/cuqi/experimental/mcmc/_utilities.py
+++ b/cuqi/experimental/mcmc/_utilities.py
@@ -6,7 +6,7 @@ import numpy as np
 import cuqi.experimental.mcmc as samplers 
 
 def find_valid_samplers(target, as_string = True):
-        """
+    """
         Finds all possible sampleras that can be used for sampling from the target distribution.
 
         Parameters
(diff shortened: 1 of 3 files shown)


jeverink:
@jakobsj @amal-ghamdi @chaozg ,
this code, as an extension of the `cuqi.experimental.mcmc.find_valid_samplers` I made some time ago, was a feature I had in my mind for a while ago and will probably write an issue for it later. But I would love to hear your feedback on this idea, it's features and possible applications, and how to improve on this.

amal-ghamdi:
Thank you @jeverink for this functionality. It seems very useful! we will take a closer look to answer your questions. We assigned ourselves for initial review.

cuqi/experimental/mcmc/_utilities.py
```diff
@@ -1,17 +1,164 @@
...
+        return None
+
+    par_names = target.get_parameter_names()
+
+    valid_samplers = dict()
+    for par_name in par_names:
+        conditional_params = {par_name_: np.ones(target.dim[i]) for i, par_name_ in enumerate(par_names) if par_name_ != par_name}
+        conditional = target(**conditional_params)
+
+        samplers = find_valid_samplers(conditional, as_string)
```
amal-ghamdi, line 66: Is this framework able to detect for example using direct samplers for conjugate variables?

cuqi/experimental/mcmc/_utilities.py
```diff
@@ -1,17 +1,164 @@
...
+        if len(samplers) == 0:
+            return None
+        
+        valid_samplers[par_name] = samplers
+
+    return valid_samplers
+
+def suggest_sampler(target, as_string = False, exceptions = []):
```
amal-ghamdi, line 74: just a slight name suggestion here, we can name it: "recommend_sampler" to make it stronger. We already have find valid samplers, since we think this is the preferred, we can use the word recommend instead of suggest

cuqi/experimental/mcmc/_utilities.py
```diff
@@ -1,17 +1,164 @@
...
+
+    valid_samplers = find_valid_samplers(target, as_string = False)
+    
+    for suggestion, values in ordering:
+        if suggestion in valid_samplers and suggestion not in exceptions:
+            # Sampler found
+            if as_string:
+                return suggestion.__name__
+            else:
+                return suggestion(target, **values)
```
amal-ghamdi, line 127: one idea here, is instead of returning the instantiated sampler, to return the class and the kwargs, to give the user flixibility to pass other parameters, like initial point, etc.

recommended_sampler_class, recomended_kwargs = recommend_sampler(target)

cuqi/experimental/mcmc/_utilities.py
```diff
@@ -1,17 +1,164 @@
...
+
+    # No sampler can be suggested
+    return None
+
+def suggest_sampling_strategy(target, as_string = False):
```
amal-ghamdi, line 132: Here I also recommend the name: recommended_HybridGibbs_sampling_strategy. Using the word recommend to indicate this is what we think is preferred, and using HybridGibbs for clarity.

cuqi/experimental/mcmc/_utilities.py
```diff
@@ -1,17 +1,164 @@
...
 
-    return valid_samplers
\ No newline at end of file
+    # Need a separate case for HybridGibbs
+    if find_valid_sampling_strategy(target) is not None:
+        valid_samplers += [cuqi.experimental.mcmc.HybridGibbs.__name__ if as_string else cuqi.experimental.mcmc.HybridGibbs]
+
+    return valid_samplers
+
+def find_valid_sampling_strategy(target, as_string = True):
```
amal-ghamdi, line 39: name suggestion: 
```suggestion
def find_valid_HybridGibbs_sampling_strategy(target, as_string = True):
```
for clarity

amal-ghamdi (commented):
Hi @jeverink, many thanks for your PR! it looks very useful building on choosing valid samplers to recommend the preferred sampler with parameter/strategy suggestion.

I made some comments throughout and I have a question below. I also wanted to ask you if you like to discuss sampler order, I will bring it up in the stand up today and we can set up a meeting for that. 

The question is how is the choice of the recommended sampler here differs from the BayesianProblem.. does the Bayesian problem simply choose the first valid sampler?

Thank you @jeverink

cuqi/experimental/mcmc/_utilities.py
```diff
@@ -1,17 +1,164 @@
...
+        Finds all possible sampleras that can be used for sampling from the target distribution.
+
+        Parameters
+        ----------
+
+        target : `cuqi.distribution.Distribution`
+            The target distribution to sample.
+        
+        as_string : boolean
+            Whether to return the name of the sampler as a string instead of instantiating a sampler. *Optional*
```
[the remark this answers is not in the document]
jeverink: Currently it instantiates, which would allows the user to immediately use the returned sampler. Would it be an idea to make this a boolean option of the method?

cuqi/experimental/mcmc/_utilities.py
jeverink: It tries to find for the conditional distribution any valid sampler, where sampler is any subclass of `cuqi.experimental.mcmc.Sampler`. As `class Direct(Sampler)`, it should therefore be able to find the Direct sampler. However as conditional distributions are generally in instance of the `Posterior`, which do not have a `sample` method that won't raise an exception, the Direct sampler won't be valid.

chaozg (commented):
@jeverink thanks for this PR. I'm just curious if it makes sense to add some tests/demos to see if the added routines indeed suggest expected samplers.

amal-ghamdi:
> Thanks for the feedback @amal-ghamdi ,
> 
> As for your question regarding how this compares to BayesianProblem. BayesianProblem currently has a list of if-else statements to pick the a sampler, with an ordering similar to the one I currently put in the `recommend_sampler` method. I think they currently will use the same samplers when provided with the same distribution.
> 
> My hope is that this utility could, at a later date, be used to replace the if-else statement chain in BayesianProblem and the all the repeated code in the `_sample*` methods in BayesianProblem, therefore making the code easier to maintain.

Many thanks @jeverink for the clarification. I took a closer look at `sample_posterior` and to my understanding, it does more than selecting the first sampler. The if-else chain also decide the sampler based on the posterior structure, while `find_valid_samplers` seems to be only used for testing to construct all possible instances of samplers for test purpose.

To conclude, it seems to me the solution you suggest here does not take into account choosing the best sampler for a particular structure of posterior like what `sample_posterior` does. And it can be confusing to users to have to different `recommenders` that acts differently, i.e. `recommend_sampler` and the recommendation from the `BayesianProblem`.  

I agree with you in general that refactoring the if-else structure might be something nice to do, but I am not sure if what we have in this PR suffice. Let me know what you think.

jeverink:
Hi @amal-ghamdi,

>  It only select all samplers that can be created without error:
is indeed what it does, and every (experimental) sampler has the `validate_target` method which is run when a target distribution is set and raises an error when the target distribution is not supported by the sampler. And it is `validate_target` that does the structure checking, i.e., `validate_target` in  `LinearRTO` checks for a linear model and Gaussian likelihood, and whether the prior Gaussian is prior. Checking the posterior structure is therefore delegated to each `Sampler` instead of the automatic sampler selection method.

I hope this clarifies it a bit.

amal-ghamdi:
> Hi @amal-ghamdi,
> 
> > It only select all samplers that can be created without error:
> 
> is indeed what it does, and every (experimental) sampler has the `validate_target` method which is run when a target distribution is set and raises an error when the target distribution is not supported by the sampler. And it is `validate_target` that does the structure checking, i.e., `validate_target` in `LinearRTO` checks for a linear model and Gaussian likelihood, and whether the prior Gaussian is prior. Checking the posterior structure is therefore delegated to each `Sampler` instead of the automatic sampler selection method.
> 
> I hope this clarifies it a bit.
if preordered samplers are:
[samplerA
samplerB,
.
.]
if we have posterior1 and posterior2 that both can be sampled by samplerA and samplerB (e.g. `validate_target` does not complain). It could be a case where samplerA is better for posterior1 and SamplerB is better for posterior2 based on some structure information (prior, posterior, model info, gradient, maybe even in the future information like having hessian, or multi scale model or surrogate model or other information like problem size, system memory, etc). The solution you suggested will always choose samplerA for both. 

That is why i think that through the if-else chain now (which can be refactored in future) we have more control to grow the recommender system and does not make it tied to preorder set of samplers.

This is why I see that the preorder set of samplers define absolute preference of samplers without taking information about the underlying problem into account.

cuqi/experimental/_recommender.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+
+    def __init__(self, target:cuqi.density.Density, exceptions = []):
+        self._target = target
+        self._exceptions = exceptions
+        self._create_ordering()
+
+    @property
+    def target(self) -> cuqi.density.Density:
+        """ Return the target density. """
```
chaozg, line 29: Distribution instead of Density?
```suggestion
        """ Return the target Distribution. """
```

cuqi/experimental/_recommender.py
```diff
@@ -0,0 +1,200 @@
...
+        return self._target
+    
+    @target.setter
+    def target(self, value:cuqi.density.Density):
+        """ Set the target density. Runs validation of the target. """
```
chaozg, line 34:
```suggestion
        """ Set the target Distribution. Runs validation of the target. """
```

chaozg (approved):
LGTM. I just left a trivial suggestion,and feel free to omit it.

amal-ghamdi (approved):
Many thanks @jeverink for the update and the restructure. I have one typo comment. Other than that I think the PR looks great.

I added issue #663 to keep track of what needed in the future, the integration with BayesianProblem. Feel free to update the issue.

PR #632, opened 2025-03-08
Fix geometry printing to account for case where geometry fun_shape is None

amal-ghamdi:


amal-ghamdi:
@jakobsj @chaozg , this is just a quick fix for geometry printing to account for case when fun_shape is None. I ran into this when trying to run the myula/poisson notebook

chaozg (approved):
LGTM

jakobsj (approved):

PR #634, opened 2025-03-14
Fix NumPy 2.0.0 support

### LINKED ISSUE #633: CUQIpy does not support Numpy 2.0.0
**Description**

Numpy version 2.0.0 removed some aliases used in CUQIpy, see `https://numpy.org/devdocs/release/2.0.0-notes.html#numpy-2-0-python-api-removals`

**Example to reproduce**

In particular, line 40 from `https://github.com/CUQI-DTU/CUQIpy/blob/main/cuqi/distribution/_truncated_normal.py`
```python
    def __init__(self, mean=None, std=None, low=-np.Inf, high=np.Inf, is_symmetric=False, **kwargs):
```

This seems to be the only instance causing issues.

jeverink:
Replace removed Numpy alias such that CUQIpy now support NumPy version 2.0.0. (Tested it with version 2.0.2, but not extensively)

Fixed #633

Perhaps consider having some tests with more recent versions of foundational packages?

### DIFF
--- a/cuqi/distribution/_truncated_normal.py
+++ b/cuqi/distribution/_truncated_normal.py
@@ -37,7 +37,7 @@ class TruncatedNormal(Distribution):
         p = cuqi.distribution.TruncatedNormal(mean=0, std=1, low=-2, high=2)
         samples = p.sample(5000)
     """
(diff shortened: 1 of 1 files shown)


nabriis (approved):
LGTM

chaozg (approved):
LGTM

PR #639, opened 2025-03-24
Add UnboundedUniform distribution

### LINKED ISSUE #638: Add non-informative/unbounded-uniform distribution
Currently, a snippet like the following works OK to draw samples based only on the likelihood:
```python
y = cuqi.distribution.Gaussian(A, noise_std**2)
data_dist = y(y=y_obs)
np.random.seed(0)
sampler_likelihood = cuqi.experimental.mcmc.NUTS(data_dist)
sampler_likelihood.sample(100)
```

However, the code might be difficult to understand as we don't explicitly define an `x` we will be sampling. As suggested by @jeverink, to be intuitive and consistent with the use of other priors, it might be better to add a non-informative/unbounded-uniform distribution.

chaozg:
fixed #638 

This PR added the proposed `UnboundedUniform `

### DIFF
--- a/cuqi/distribution/__init__.py
+++ b/cuqi/distribution/__init__.py
@@ -14,6 +14,6 @@ from ._lognormal import Lognormal
 from ._normal import Normal
 from ._truncated_normal import TruncatedNormal
 from ._posterior import Posterior
-from ._uniform import Uniform
+from ._uniform import Uniform, UnboundedUniform
 from ._custom import UserDefinedDistribution, DistributionGallery
 from ._joint_distribution import JointDistribution, _StackedJointDistribution, MultipleLikelihoodPosterior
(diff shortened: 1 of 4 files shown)


chaozg:
> Thanks for the feature @chaozg , Perhaps you could add a test that you indeed cannot sample from the UnboundedUniform?

I just added a unit test for this purpose and now request your further review.

jeverink (approved):
LGTM

jakobsj:
Thanks @chaozg, it is looking good. An elegant solution to subclass UnboundedUniform from Uniform, however I see the logpdf value of Uniform was modified to be unnormalised, probably in order for it to work when inherited. But isn't it undesirable that Uniform itself has its behaviour changed, and to provide only an unnormalised value when we have easy access to the normalised value with the existing implementation?

chaozg:
Hi @jakobsj , I see your concern. Indeed, the normalized pdf might be useful in some special we can't predict now, and it's better to leave `Uniform` untouched. I just subclassed `UnboundedUniform` from `Distribution` and added separate unit tests for it.

PR #660, opened 2025-06-20
Fix hybrid gibbs ordering

### LINKED ISSUE #441: Define sampling order in Gibbs
See e.g. #424

### LINKED ISSUE #657: Ordering of HybridGibbs
**Description**

The order in which the HybridGibbs samples from the conditional distributions is currently the order in which they are combined in the JointDistribution. This could cause confusion when providing user-specified initial-guesses for some variables, as these initial-guesses might not be used at all, impacting the results.

It would perhaps be more natural to use the ordering in which the sampling strategy is specified instead, such the ordering is explicitly described as part of the sampler, not as part of the distribution.

**Example to reproduce**

Image

Specific problematic code from the HybridGibbs class:
Image
Image

jeverink:
Changes the ordering of the sampler in `HybridGibbs` to the order of the sampling_strategy instead of the order in which the posterior was constructed.

Closes #657 

With scan_order:
Closes #441

### DIFF
--- a/cuqi/experimental/mcmc/_gibbs.py
+++ b/cuqi/experimental/mcmc/_gibbs.py
@@ -134,7 +134,7 @@ class HybridGibbs:
         self._scan_order = scan_order
 
         # Check that the parameters of the target align with the sampling_strategy and scan_order
-        if set(self.par_names) != set(self._scan_order):
+        if set(self.par_names) != set(self.scan_order):
             raise ValueError("Parameter names in JointDistribution do not equal the names in the scan order.")
 
         # Initialize sampler (after target is set)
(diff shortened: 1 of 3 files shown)


chaozg (approved):
LGTM 👍 and thanks for adding the explicit statement on sampling order.

cuqi/experimental/mcmc/_gibbs.py
```diff
@@ -121,6 +129,14 @@ def __init__(self, target: JointDistribution, sampling_strategy: Dict[str, Sampl
         # Store parameter names
         self.par_names = self.target.get_parameter_names()
 
+        # Store the scan order
+        self._scan_order = scan_order
+
+        # Check that the parameters of the target align with the sampling_strategy and scan_order
+        for par_name in self.par_names:
+            if par_name not in self.scan_order:
```
amal-ghamdi, line 137: can may be use condition like
```
if set(self.par_names) != set(self._scan_order):
    #raise Error
```
to ensure scan order is not only a superset, but equal to the par_names

PR #662, opened 2025-06-20
Print for HybridGibbs

### LINKED ISSUE #629: Printing (__repr__) of HybridGibbs
Currently, `cuqi.experimental.mcmc.HybridGibbs` is the only sampler without a nice printing representation (`__repr__`), due to it not inheriting from  `cuqi.experimental.mcmc.Sampler`.

jeverink:
Added `__repr__` for `HybridGibbs`.

Close #629

### DIFF
--- a/cuqi/experimental/mcmc/_gibbs.py
+++ b/cuqi/experimental/mcmc/_gibbs.py
@@ -338,11 +338,11 @@ class HybridGibbs:
         """ Return a string representation of the sampler. """
         msg = f"Sampler: {self.__class__.__name__} \n"
         if self.target is None:
-            msg += f"Target: None \n"
+            msg += f" Target: None \n"
         else:
-            msg += f"Target: \n \t {self.target} \n\n"
+            msg += f" Target: \n \t {self.target} \n\n"
             
         for key, value in zip(self.samplers.keys(), self.samplers.values()):
-            msg += f"Variable '{key}' with {value} \n"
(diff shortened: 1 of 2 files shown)


cuqi/experimental/mcmc/_gibbs.py
```diff
@@ -333,3 +333,16 @@ def _store_samples(self):
...
+
+    def __repr__(self):
+        """ Return a string representation of the sampler. """
+        msg = f"Sampler: {self.__class__.__name__} \n"
+        if self.target is None:
+            msg += f"Target: None \n"
+        else:
+            msg += f"Target: \n \t {self.target} \n\n"
+            
+        for key, value in zip(self.samplers.keys(), self.samplers.values()):
```
chaozg, line 345: I'm just curious if it is needed/preferred to check subsampler's `._is_initialized:` first, as in the base [Sampler](https://github.com/CUQI-DTU/CUQIpy/blob/main/cuqi/experimental/mcmc/_sampler.py) class.

chaozg (approved):
Approved (with a tiny question)

cuqi/experimental/mcmc/_gibbs.py
jeverink: The `value` is a sampler instance, which when formatted uses that sampler's `__repr__`.
So it looks something like:
image

PR #667, opened 2025-06-30
Allow other options for initial guess in (Regularilzed)LinearRTO steps

chaozg:
Retired in favour of #679 

## Do solvers accept 0 as the maximum number of iterations?
- CUQIpy's CGLS: yes
- CUQIpy's FISTA: yes (updated in this PR)
- CUQIpy's ADMM: yes (updated in this PR)
- [scipy.optimize.minimize](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html) through `maxiter`: no (a maxiter of 0 is not honored)
- [scipy.optimize.lsq_linear](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.lsq_linear.html#scipy.optimize.lsq_linear) through `max_iter`: no (error raised)

##

### DIFF
--- a/cuqi/experimental/mcmc/_rto.py
+++ b/cuqi/experimental/mcmc/_rto.py
@@ -52,23 +52,21 @@ class LinearRTO(Sampler):
         # Other parameters
         self.maxit = maxit
         self.tol = tol
-        self.specified_inner_initial_point = inner_initial_point
+        self.inner_initial_point = inner_initial_point
 
     def _initialize(self):
         self._precompute()
-        self._set_inner_initial_point()
-
-    def _set_inner_initial_point(self):
-        if self.specified_inner_initial_point == "previous_sample":
-            self._inner_initial_point = self.current_point
-        elif self.specified_inner_initial_point == "MAP":
-            self._inner_initial_point = self._map
-        else:
-            raise ValueError("inner_initial_point must be either 'previous_sample' or 'MAP'.")
 
     @property
     def inner_initial_point(self):
         return self._inner_initial_point
+    
+    @inner_initial_point.setter
+    def inner_initial_point(self, value):
+        if value == "previous_sample" or value == "MAP":
+            self._inner_initial_point = value
+        else:
+            raise ValueError("inner_initial_point must be either 'previous_sample' or 'MAP'.")
 
     @property
     def prior(self):
@@ -136,7 +134,13 @@ class LinearRTO(Sampler):
 
     def step(self):
         y = self.b_tild + np.random.randn(len(self.b_tild))
-        sim = CGLS(self.M, y, self.inner_initial_point, self.maxit, self.tol)
+        if self.inner_initial_point == "MAP":
+            sim = CGLS(self.M, y, self.map, self.maxit, self.tol)
+        elif self.inner_initial_point == "previous_sample":
+            sim = CGLS(self.M, y, self.current_point, self.maxit, self.tol)
+        else:
+            # This should never happen, as we check the value of inner_initial_point in the setter.
+            raise ValueError("inner_initial_point must be either 'previous_sample' or 'MAP'.")
         self.current_point, _ = sim.solve()
         acc = 1
         return acc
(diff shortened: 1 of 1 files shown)


cuqi/experimental/mcmc/_rto.py
```diff
@@ -36,18 +36,24 @@ class LinearRTO(Sampler):
...
+        If True, the initial point for the inner CGLS solver is set to the MAP estimate of the unperturbed linear least squares problem, and the samples will be independent.
+        If False, the initial point for the inner CGLS solver is set to the current point of the sampler, and the samples will be dependent.
+
     callback : callable, optional
         A function that will be called after each sampling step. It can be useful for monitoring the sampler during sampling.
         The function should take three arguments: the sampler object, the index of the current sampling step, the total number of requested samples. The last two arguments are integers. An example of the callback function signature is: `callback(sampler, sample_index, num_of_samples)`.
         
     """
-    def __init__(self, target=None, initial_point=None, maxit=10, tol=1e-6, **kwargs):
+    def __init__(self, target=None, initial_point=None, maxit=10, tol=1e-6, independent=False, **kwargs):
```
amal-ghamdi, line 49: I suggest renaming independent to something more explicit. For example: "inner_initial_point". This also works well with the naming convention in RegularizedRTO. For example, we have `maxit` and `inner_max_it`, `abstol` and `inner_abstol`. Then we will have `initial_point` and `inner_initial_point`.
The values of `inner_initial_point` can then be "map" or "previous_sample", for example.

cuqi/experimental/mcmc/_rto.py
```diff
@@ -111,10 +117,17 @@ def M(x, flag):
             self.M = M  
         else:
             raise TypeError("All likelihoods need to be callable or none need to be callable.")
+        # compute the map
+        sim = CGLS(self.M, self.b_tild, self.current_point, self.maxit, self.tol)            
+        self.map, _ = sim.solve()
```
amal-ghamdi, line 122: I suggest making map a hidden variable `self._map` because it is not standard in our samplers and we have a custom approach to compute the map for now (through Bayesian problem)

cuqi/experimental/mcmc/_rto.py
```diff
@@ -228,6 +246,8 @@ def _initialize(self):
             self.solver = "FISTA" if callable(self.proximal) else "ADMM"
         if self.solver == "FISTA":
             self._stepsize = self._choose_stepsize()
+        self.compute_map_regularized()
+        self.initial_point_per_step = None
```
amal-ghamdi, line 250: I also suggest these two to be hidden
```suggestion
        self._compute_map_regularized()
        self._initial_point_per_step = None
```

cuqi/experimental/mcmc/_rto.py
```diff
@@ -272,15 +292,55 @@ def _choose_stepsize(self):
...
+            # Adapt bounds format, as scipy.minimize requires a bounds format 
+            # different than that in scipy.lsq_linear.
+            bounds = [(self.target.prior._box_bounds[0][i], self.target.prior._box_bounds[1][i]) for i in range(self.target.prior.dim)]
+            # Note that the objective function is defined as 0.5*||Mx-y||^2, 
+            # and the corresponding gradient (gradfunc) is given by M^T(Mx-y).
+            sim = ScipyMinimizer(lambda x: 0.5*np.sum((self.M(x, 1)-y)**2), self.initial_point, gradfunc=lambda x: self.M(self.M(x, 1) - y, 2), bounds=bounds, tol=self.abstol, options={"maxiter": self.maxit})
+        else:
+            raise ValueError("Choice of solver not supported.")
+
+        self.map_regularized, _ = sim.solve()
```
amal-ghamdi, line 331: And this one as well, also we can call it `_map` for consistency with RTO
```suggestion
        self._map_regularized, _ = sim.solve()
```

cuqi/experimental/mcmc/_rto.py
```diff
@@ -272,15 +292,55 @@ def _choose_stepsize(self):
...
+
     def step(self):
```
amal-ghamdi, line 333: to reduce code redundancy and enhance maintainability, I suggest keeping the logic that solves for point (map or sample) in one place. It can be called `_customized_step(pertrub, ...)` for example.

Then define `step` and `_compute_map_regularized` as follows
```
def step():
    _customized_step(True, ...)

def _compute_map_regularized():
    _customized_step(False, ...)
```
_customized_step behave differently based on its input, for example, calling it through step should save the results in current_point, but calling it through  `_compute_map_regularized` should update the `_map`

amal-ghamdi:
It will also be great to add some tests.

PR #675, opened 2025-07-27
Array agnostic refactor via cursor

nabriis:


cuqi/array/_array.py
```diff
@@ -1,108 +1,266 @@
-import numpy as np
 from cuqi.geometry import _DefaultGeometry1D
 
+# Import the backend module directly to avoid circular import
+import os
+_backend_name = os.getenv("CUQI_ARRAY_BACKEND", "numpy").lower()
```
nabriis, line 5: It would be good if users could set the backend diretly in the script using e.g. cuqi.array.set_backend("pytorch") or similar. This may mean rethinking the class creation idea for handing the CUQIarray.

cuqi/distribution/_gaussian.py
```diff
@@ -1,7 +1,8 @@
 import warnings
 import numbers
-import numpy as np
-import numpy.linalg as nplinalg
+import cuqi.array as xp
```
nabriis, line 3: two imports here.

cuqi/geometry/_geometry.py
```diff
@@ -981,21 +980,21 @@ def __init__(self, grid, std=1.0, cor_len=0.2, nu=3.0, axis_labels=None, **kwarg
...
 
     # computes the real function out of expansion coefs
     def par2fun(self,p):
-        freq = np.zeros(self.par_dim)
+        freq = xp.zeros(self.par_dim)
         m = len(p)
         freq[:m] = p
         temp = freq*self.coefs
-        real = idst(temp)/2/np.pi
+        real = idst(temp)/2/xp.pi
```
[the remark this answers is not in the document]
nabriis: Don't do this.

cuqi/solver/_solver.py
```diff
@@ -1,12 +1,12 @@
-import numpy as np
-from numpy import linalg as LA
+import cuqi.array as xp
+LA = xp.linalg  # Linear algebra operations
 from scipy.optimize import fmin_l_bfgs_b, least_squares
 import scipy.optimize as opt
```
nabriis, line 4: For scipy. an we add this under the xp also? xp.optimize and xp.sparse? Add NotImplementedError if not implemented in e.g. PyTorch.

demos/howtos/1d_deconvolution_mrf_priors.py
```diff
@@ -0,0 +1,542 @@
...
+"""
+
+import os
+import sys
+import numpy as np
+import matplotlib.pyplot as plt
+from contextlib import contextmanager
+
+@contextmanager
+def backend_context(backend_name):
```
nabriis, line 27: Please make this whole howto much simpler and do not use this many functions. Instead make it script-like. Showcase and plit using the built-in plotting in CUQIpy. Switch between Numpy and PyTorch to show it gives the same results.

cuqi/utilities/_utilities.py
```diff
@@ -12,15 +12,16 @@
...
+            value = xp.array(value)
         else:
-            value = np.array(value).reshape((1,1))
+            value = xp.array(value).reshape((1,1))
             
         if flatten is True:
             value = value.flatten()
-    if isinstance(value,np.matrix): #Convert to array if matrix (matrix acts different on (n,) arrays)
+    # Convert numpy matrix to array if needed (matrix acts different on (n,) arrays)
+    if hasattr(value, 'A') and hasattr(value, 'shape') and callable(getattr(value, 'A', None)):
```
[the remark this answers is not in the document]
nabriis: Seems like not a good approach. Instead I suggesting checking backend, and if NumPy use the same check as before.

nabriis (commented):
1) Please remove the image 1d_deconvolution_mrf_comparsion.png. We do not want this added in the root
2) Please refactor and rewrite the two added .md files in root. Instead update the documentation under docs/ to reflect these changes. There are a number of areas where this would need to happen. Add most technical details in the technical area.
3) Check my review comments and take them into account.

nabriis:
Also unit tests are now failing. Run locally first to ensure they pass.

PR #679, opened 2025-08-21
Allow other options for initial guess in (Regularilzed)LinearRTO steps

### LINKED ISSUE #653: Initial guess for solves in RegularizaedRTO
The current implementation of the sampling process in RegularizaedRTO makes the samples Markov, if the solver is not fully converged.

### LINKED ISSUE #684: Add getter and setter for current_point
Add getter and setter for sampler's `current_point`

This is really a tiny issue, and I created it solely for documentation and traceability purposes.

chaozg:
Fixed #653 
Fixed #684

### DIFF
--- a/cuqi/experimental/mcmc/_rto.py
+++ b/cuqi/experimental/mcmc/_rto.py
@@ -51,7 +51,7 @@ class LinearRTO(Sampler):
         # Other parameters
         self.maxit = maxit
         self.tol = tol
-        self._specified_inner_initial_point = inner_initial_point
+        self.inner_initial_point = inner_initial_point
 
     def _initialize(self):
         self._precompute()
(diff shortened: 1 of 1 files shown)


cuqi/experimental/mcmc/_rto.py
```diff
@@ -36,21 +36,40 @@ class LinearRTO(Sampler):
...
+            if self.specified_inner_initial_point == "previous_sample":
+                return self.current_point
+            elif self.specified_inner_initial_point == "MAP":
+                return self._map
+            else:
+                raise ValueError("Invalid value for inner_initial_point. Choose either 'previous_sample', 'MAP', or provide a numpy array/cuqi array.")
+        elif isinstance(self.specified_inner_initial_point, (np.ndarray, cuqi.array.CUQIarray)):
+            return self.specified_inner_initial_point
+        else:
+            raise ValueError("Invalid value for inner_initial_point. Choose either 'previous_sample', 'MAP', or provide a numpy array/cuqi array.")
```
amal-ghamdi, line 72: The getter logic here has some setter logic in my view (detecting the errors). Maybe it can be splitted to setter and getter:
```python
    @property
    def inner_initial_point(self):
        if isinstance(self.specified_inner_initial_point, str):
            if self.specified_inner_initial_point == "previous_sample":
                return self.current_point
            elif self.specified_inner_initial_point == "map":
                return self._map
        else:
            return self.specified_inner_initial_point


    @inner_initial_point.setter
    def inner_initial_point(self, value):
        is_correct_string = (isinstance(value, str) and
                             (value.lower() == "previous_sample" or
                              value.lower() == "map"))
        if is_correct_string:
            self.specified_inner_initial_point = value.lower()
        elif isinstance(value, (np.ndarray, cuqi.array.CUQIarray)):
            self.specified_inner_initial_point = value
        else:
            raise ValueError("Invalid value for inner_initial_point. Choose either 'previous_sample', 'MAP', or provide a numpy array/cuqi array.")
```

And then in the initializer here and in the RegularizedLinearRTO, we can just use the setter
```python
self.inner_initial_point = inner_initial_point
```

PR #681, opened 2025-08-27
Add TweediePrior and ScorePrior as alias of MoreauYoshidaPrior

### LINKED ISSUE #670: Add ScorePrior to be used with PnPULA

chaozg:
fixed #670

### DIFF
--- a/cuqi/implicitprior/_restorator.py
+++ b/cuqi/implicitprior/_restorator.py
@@ -237,17 +237,19 @@ class MoreauYoshidaPrior(Distribution):
 class TweediePrior(MoreauYoshidaPrior):
     """
     Alias for MoreauYoshidaPrior following Tweedie's formula framework. TweediePrior 
-    defines priors where gradients are computed using MMSE (Minimum Mean Square Error) denoisers.
+    defines priors where gradients are computed based on Tweedie's identity that links 
+    MMSE (Minimum Mean Square Error) denoisers with the underlying smoothed prior, see:
+        - Laumont et al. https://arxiv.org/abs/2103.04715 or https://doi.org/10.1137/21M1406349
 
     Tweedie's Formula
     -------------------------
-    In the context of denoising, Tweedie's formula states that for a signal x
+    In the context of denoising, Tweedie's identity states that for a signal x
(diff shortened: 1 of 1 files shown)


cuqi/implicitprior/_restorator.py
```diff
@@ -232,4 +232,62 @@ def get_conditioning_variables(self):
...
+        Prior of the RestorationPrior type containing a denoiser/restorator.
+        
+    smoothing_strength : float, default=0.1
+        Corresponds to the noise level σ² in Tweedie's formula context.
+
+    See MoreauYoshidaPrior for the underlying implementation with complete documentation.
+    """
+    pass
+
+class ScorePrior(MoreauYoshidaPrior):
```
amal-ghamdi, line 272: I am thinking if we should link score to Tweedie here, same as, for example, in https://deepinv.github.io/deepinv/api/stubs/deepinv.optim.ScorePrior.html#deepinv.optim.ScorePrior.score


    """
    Alias for TweediePrior (and MoreauYoshidaPrior) following the score-based framework.

    ScorePrior defines (implicit) prior where the score (gradient of the log density of
    the underlying implicit prior) is computed based on Tweedie's identity.

    At implementation level, ScorePrior shares identical functionality with 
    TweediePrior (and MoreauYoshidaPrior), and it is implemented as an alias
    of them. The separate name provides clarity when working
    specifically with score-based approaches.

    Parameters
    ----------
    prior : RestorationPrior
        Prior of the RestorationPrior type containing a denoiser/restorator.

    smoothing_strength : float, default=0.1
        Smoothing strength for the score-based prior (corresponding to the noise level
        in epsilon in TweediePrior).

    See TweediePrior and MoreauYoshidaPrior for further documentation and the underlying
    implementation.
    """

amal-ghamdi (commented):
Thank you @chaozg for this PR. I think it is reasonable to implement these two classes as alias. I made few suggestions.

I am also open to keeping or removing the ScorePrior altogether. In some sense, it is good to have it visible for users familiar with this term, on the other hand, the documentation is a bit less rigorous compared to MoreauYoshidaPrior and TweediePrior classes.

PR #683, opened 2025-09-10
Ensure NUTS statefulness within Gibbs

### LINKED ISSUE #512: Make NUTS fully stateful in HybridGibbs

### LINKED ISSUE #621: Improve statefulness of using NUTS within Gibbs

amal-ghamdi:
closes #512
closes #621

### DIFF
--- a/tests/zexperimental/test_mcmc.py
+++ b/tests/zexperimental/test_mcmc.py
@@ -1641,14 +1641,18 @@ def test_NUTS_within_Gibbs_consistant_with_NUTS(step_size, num_sampling_steps_x,
     sampler_nuts = cuqi.experimental.mcmc.NUTS(target,
                                                max_depth=4,
                                                step_size=step_size)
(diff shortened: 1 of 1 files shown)


nabriis (approved):
Huge! Glad it seems to be fully statefull now! 🤘

PR #690, opened 2025-09-19
add online thinning option to samplers

### LINKED ISSUE #677: Online thinning of samples

In image-related problems, we may need to run long chains of millions of steps. Storing all intermediate samples is computationally expensive and often unnecessary, so implementing online thinning would be beneficial.

**Describe the solution you'd like**

```python
Ns=100
Nt=10 # Nt is thinning interval/frequency
sampler.sample(Ns, Nt)
```

Related to #691

chaozg:
fixed #677

### DIFF
--- a/cuqi/experimental/mcmc/_sampler.py
+++ b/cuqi/experimental/mcmc/_sampler.py
@@ -203,7 +203,7 @@ class Sampler(ABC):
 
         self.set_state(state)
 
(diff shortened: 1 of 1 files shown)


amal-ghamdi (approved):
Thanks @chaozg, LGTM!

jakobsj (approved):
Looks great, always a good sign when a new feature can be permitted with very few lines of code. Very nice test as well using resetting of the random seed to check equivalence of samples stored using different thinning levels.

